PR #46, opened 2020-04-19
Fix infinite requests being sent to the API

arku:


katelovescode:
So this is a particularly frustrating bit for me, I've tried a whole bunch of ways to solve the linting issue this presents, but I can't seem to get it.  So can you add a `// eslint-disable-next-line react-hooks/exhaustive-deps` before the dependency array in both these files?

PR #48, opened 2020-04-19
Setup react-query

arku:
# Before You Submit:
 * [ ] Did you write tests?

## 📋 Summary

Uses [react-query](https://github.com/tannerlinsley/react-query) to clean up asynchronous data fetching.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -8,15 +8,16 @@
   },
   "proxy": "http://localhost:3001",
   "dependencies": {
+    "axios": "^0.19.2",
     "csv": "^5.3.2",
     "hash-anything": "^1.2.2",
     "prop-types": "^15.7.2",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "react-ga": "^2.7.0",
+    "react-query": "^1.2.7",
     "react-router-dom": "^5.1.2",
     "react-scripts": "^3.4.1",
-    "react-use-fetch-api": "^1.0.8",
     "uuid": "^7.0.3",
     "xlsx": "^0.15.6"
   },
--- a/client/src/Login/Login.js
+++ b/client/src/Login/Login.js
@@ -1,9 +1,10 @@
-import React, { useState, useEffect } from 'react'
+import React from 'react'
 import { NavLink } from 'react-router-dom'
 import './Login.css'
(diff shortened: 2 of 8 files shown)


katelovescode:
Can you talk through some of the justification for this choice?  I'm not sure why we'd need to change what we're using currently.

arku:
@katelovescode Sure!

The functionality of `react-use-fetch-api` is quite limited IMHO. It lets us make AJAX requests using the `fetch` API. That's it.

**Cons**:

- No easy way to cache the data (to avoid re-fetching the data every time) and to re-fetch the data when necessary.

-  Error handling is not straightforward from what I see. We can pass an `onError` function but this requires us to manage the `error` state. How about status checks? Again, we need to maintain our own `loading` or `status` piece of state. More code to write!

Not to mention the pesky infinite requests bug 🙂 

On the other hand, `react-query` solves both the above issues with us [writing less code](https://github.com/pieforproviders/pieforproviders/pull/48/files#diff-30763305b61fee9f9e7785a8c6781f4cR16) (less is better right?). It has a simple API that lets us access the status, data, and errors if any of every query. It also manages a query cache and does auto caching and data re-fetching.

Let me know if you have any questions.

katelovescode:
Definitely get your points.  However, I'd prefer to stick with what we have for a few reasons:

1. This is an MVP and I won't have a better idea of what we're building out until I get a stronger sense of product requirements.  I don't want to overoptimize before I figure out the shape and frequency of the API calls we'll be making (obviously the infinite loop is a problem, but it has an easy solution re: not listing dependencies).

2. We have control of the API we're calling; we're responsible for hosting it and it's built on strict JSON API definitions; that doesn't mean error handling isn't important on the front-end, but it does mean we have access to diagnose the problem on the back end if there is one.  `react-query` is a _heavy_ dependency and I don't want to hide interaction between the API and front-end behind a large library; partially because this app is also a learning opportunity for me, and I don't want to abstract away functionality before I fully understand it.

Hopefully that makes sense - if you're cool with it, I'd love for you to make an issue for upgrading to react-query and I'll punt it til after MVP.

arku:
@katelovescode Got it. Closing this for now.

PR #59, opened 2020-04-19
feat: Setup Sentry to track error events

### LINKED ISSUE #47: Add Sentry for bugcatching

arku:
# Before You Submit:
 * [ ] Did you write tests?

## Deployment Considerations

We need to set an environment variable `REACT_APP_SENTRY_DSN` before building the code for Production.

## 📋 Summary

Sets up [Sentry](https://sentry.io) to track error events in Production. Fixes #47

When a user encounters an error, they get to share feedback that gets sent to Sentry along with the event information.

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ coverage
 # Dotenv
 **/*.env*
 !.env.sample
+!.env.development.sample
 
 # for rvm
 .ruby-gemset
--- a/client/.env.development.sample
+++ b/client/.env.development.sample
@@ -1,0 +1,1 @@
+REACT_APP_SENTRY_DSN=[leave empty]
--- a/README.md
+++ b/README.md
@@ -63,7 +63,8 @@ Learn more at [www.pieforproviders.com](http://www.pieforproviders.com)
 
 - clone the repo
 - `cd pieforproviders`
-- copy `.env.sample` to `.env` and add values (contact a repo contributor)
+- Run `cp .env.sample .env` and `cp client/.env.development.sample cp client/.env.development` and add values (contact a repo contributor)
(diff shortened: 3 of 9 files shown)


.gitignore
```diff
@@ -30,9 +30,10 @@ coverage
 # Dotenv
 **/*.env*
 !.env.sample
+!.env.development.sample
```
katelovescode, line 33: Is there a reason to keep this in the .development .env rather than the root one, given that we also need this env var on Prod?  I know it's a manual process to add it there, I just mean for documentation purposes, seems unclear that it would be in development rather than all envs (or just production for that matter.  Thoughts?

arku: Good question. We don't want to enable Sentry on development, do we? That'll be quite a lot of noise. I was thinking of having the required variables for Production in `.env` and override them locally with `.env.development` (`.env.local` will work too). That's why the code checks if the environment is `production` and if `process.env.REACT_APP_SENTRY_DSN` is set. 

Does that make sense?

katelovescode: Ah, yeah, perfect.  Good plan!

client/.env.development.sample
```diff
@@ -0,0 +1 @@
+REACT_APP_SENTRY_DSN=
```
katelovescode, line 1:
```suggestion
REACT_APP_SENTRY_DSN=[leave empty]
```

.gitignore
katelovescode: Can you add a comment to the readme that both .env.sample and .env.development will need to be copied during setup?

arku: Done.

PR #63, opened 2020-04-20
Move side effects into effects and enable GA only in Production

arku:
* [ ] Did you write tests?

Does it what it says on the tin 🙂


client/src/Login/__tests__/Login.test.js
```diff
@@ -36,14 +35,6 @@ describe('<Login />', () => {
       wrapper = shallow(<Login />)
     })
 
-    it('calls ReactGA.pageview()', () => {
```
arku, line 39: @katelovescode We could add these tests back but we have to enable ReactGA on `test` mode as well. Let me know what you think.

katelovescode: I think it's fine to remove them.

Issue #84, opened 2020-04-26
Upload .xls and .xlsx


rebeccakarasiktw:
##  🕷Description

As a user I want to be able to upload .xls and .xlsx (in addition to .csv)

## 🍭 Steps to Reproduce
1. Go to 'https://pie-staging.herokuapp.com/cases/import'
2. Click on 'CSV import'
3. Try to upload a .xls

## 🍣 Additional context

- Devices you saw the bug on: Mac laptop
- Any other information that's relevant: We expect users to have .xls files not only .csv

## 🧮 Priority

Medium

jessehall3:
Current questions:

1. For xlsx/workbooks, would the user need an option to select a particular tab or sheet? Is this looking too far ahead, are we just expecting to choose the first sheet available?

1. Any concerns with using [xlsx](https://www.npmjs.com/package/xlsx) as the library for this? It's big, but very popular and well supported.

1. Are we ok with just a simple upload button with little to no error handling and feedback for now, or are we looking to make use of a robust upload or dropzone library?

1. What should the output data look like? Right now it looks like:
`[
    [Harry, Potter, 123-45-6789],
    [Pie, Provider, 987-65-4321]
]`
Is that okay for now?

1. Are we looking to output (to the view) anything beyond a simple table for now?

katelovescode:
Hey Jesse - I think I answered these questions poorly last time we talked, I'm not sure I understood the correlation between these questions and the shifting requirements. Once you started working on this we developed a template XLS; sorry that happened while you were working on it.  A couple of ideas: we can merge your PR and write another ticket for the requirements of the new spreadsheet, or we can change the requirements of this ticket and you can either open a new PR or reopen this one and make changes.  @jessehall3 any preference about which route we take?

jessehall3:
Thanks for the update, @katelovescode  Please feel free to merge the PR and/or make a new issue as you see fit. I can't commit to working on the future issue right now but I will take a look at it if it is created.

katelovescode:
Sounds good - since we've been getting better at defining issues and reducing scope on them, we've got a couple smaller tickets if you're still interested in working on some stuff!  Sorry this one got away from us.

csprayregen:
- [ ]  .csv, 

- [ ] .xlsx

- [ ] .xls

csprayregen:
@arku @katelovescode .xls produced this error (Github doesn't support it, so I can't attach the file, but can share over slack!)

[...]

PR #104, opened 2020-05-12
Issue 84 Upload xls files

jessehall3:
#84 
Allows users to upload cases using .xls and .xlsx formatted files, in addition to csv.

# 🍂 Before You Submit:

 * [ ] Did you write tests?
I used existing tests, these are sub-par in coverage as this feature is not complete.

# 🛷 Deployment Considerations

* [ ] Data Migrations
* [ ] Schema Migrations
* [x] Dependencies
This PR adds a new dependency, [SheetJS js-xlsx](https://www.npmjs.com/package/xlsx). The library [has a pro edition](https://sheetjs.com/pro) for added performance and/or features.

# 🧵 Steps to set up locally
## Add the new library
1. `cd client`
1. `yarn install`

# 🧳 Steps to test
1. Get the app up and running.
1. Go to the csv import page: http://localhost:3000/csv-import
1. Upload an .xls file.
1. Upload an .xlsx file.
1. Observe that the data is read in these formats.

# 🕯 Caveats, concerns
1. For workbooks uploaded via xlsx, we assume the desired data is on the first spreadsheet.
1. I added an id to each kid before displaying the data. I did this because the sample files I used were creating duplicate keys because not all rows of the spreadsheets were unique (the previous keys used a hash of each row). While the rows should be unique as they contain SSN's, I was keeping in mind the idea that folks are just going to want to see things work when uploading random files at this phase.
1. A new library is introduced: [SheetJS js-xlsx](https://www.npmjs.com/package/xlsx)

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -13,14 +13,13 @@
     "hash-anything": "^1.2.2",
     "prop-types": "^15.7.2",
     "react": "^16.12.0",
-    "react-csv-reader": "^3.0.5",
     "react-dom": "^16.12.0",
     "react-ga": "^2.7.0",
     "react-router-dom": "^5.1.2",
     "react-scripts": "^3.4.1",
     "react-use-fetch-api": "^1.0.8",
     "uuid": "^7.0.3",
-    "xlsx": "^0.15.6"
+    "xlsx": "^0.16.0"
   },
   "scripts": {
     "start": "react-scripts start",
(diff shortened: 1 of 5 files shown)


katelovescode:
Sorry it took me so long to get to this, @jessehall3 

I'm seeing some weird behavior - it seems to only be importing the first three columns.

Screen Shot 2020-05-25 at 1 50 46 PM

Screen Shot 2020-05-25 at 1 50 18 PM
 I would expect to see the other columns and the data past column C on the displayed page.  Can you take a look at this?

jessehall3:
It looks like the requirements have shifted a bit since submitting this PR, which I tried to outline with my [previous questions in the issue](https://github.com/pieforproviders/pieforproviders/issues/84#issuecomment-627645104). I think I'll close out this PR for now until I can get more clarity on this.

katelovescode:
@all-contributors please add @jessehall3 for code

PR #107, opened 2020-05-16
[54] Users can sign up (UI Only)

katelovescode:

This is the first PR for issue 54 - users can sign up.  This PR only adds UI components for the signup page, along with responsive styling.

## 🍂 Before You Submit

* [ ] Did you write tests? - **No, will add tests in another PR**

* [X] Dependencies - ** yarn install**

## 🧵 Steps to set up locally
Stop your frontend server and restart it (or stop heroku local if you're running both)

## 🧳 Steps to test
Visit `localhost:3000/signup`.

This PR handles the following acceptance criteria:

- [x] I can enter all the fields present in the UX mockup
- [x] The field "Preferred Language" defaults to English
- [x] The "terms of service" phrase should link to the TOS content in a separate page

## 🕯 Caveats, concerns

Next steps:
- front-end tests
- form validation
- change the columns in the User table to `phone_type` and `phone_number` - for MVP let's work on handling one type of phone since that's all we're asking for on signup
- submit data to API


client/tailwind.config.js
```diff
@@ -0,0 +1,32 @@
+module.exports = {
```
mdbiscan, line 1: Even though you're using purgecss, you might want to look at the final stylesheet. Tailwind notoriously creates giant sizes, because by default everything is on. Either purgecss is doing a lot of heavy lifting in the post processing, because their probably is a lot of unused stylesheets. Regards, I recommend whitelisting what you need from tailwind so that it is not generating larger files than you need it to.

client/tailwind.config.js
```diff
@@ -0,0 +1,32 @@
...
+        gray2: '#4F4F4F',
+        gray3: '#828282',
+        lightGray: '#979797',
+        red1: '#D64B3A',
+        red2: '#F9E4E1',
+        orange1: '#F8921F',
+        orange2: '#FAE7D1',
+        green1: '#179D57',
+        green2: '#DCFAEA',
+        blueOverlay: 'rgba(0,74,110,0.5)'
```
mdbiscan, line 22: A good color naming patter is to spread them out by 100s, from 100 (lightest) to 900 (darkest).

For aplha colors, you can use hex/hexa colors to stay consisant, or use rgb/rgba.

https://css-tricks.com/8-digit-hex-codes

```javascript
colors: {
  blue-100: '#004A6E',
  blue-100-a-50: `#004A6E80`,
}
```

client/tailwind.config.js
```diff
@@ -0,0 +1,32 @@
...
+      }
+    }
+  },
+  plugins: [
+    // components
+    require('./tailwind/plugins/components/required')()
```
mdbiscan, line 28: :+1: 

'required' could be `../tailwind/plugins/utilities/required`

client/tailwind.config.js
katelovescode: Can you say more about this? Do you mean that I shouldn't use `extend` but instead should overwrite and only define styles I need?

client/tailwind.config.js
katelovescode: See my comment in slack (but I should definitely move to 8-digit hex codes)

client/tailwind.config.js
katelovescode: Even though I'm using `addComponent`?  How can I tell what should be a component and what should be a utility?

client/tailwind.config.js
mdbiscan: More about explicitly stating tailwind plugins and variants and the issues with tailwind file sizes. https://tailwindcss.com/docs/controlling-file-size/

client/src/Signup/Signup.js
```diff
@@ -0,0 +1,252 @@
...
+      ReactGA.event({
+        category: 'Guest',
+        action: 'Landed on Signup Page'
+      })
+    }
+  }, [])
+
+  const TermsLabel = () => {
+    return (
+      <div>
```
mdbiscan, line 41: Use a fragment instead of a `div`.

client/src/Signup/Signup.js
```diff
@@ -0,0 +1,252 @@
...
+                label: 'Español',
+                value: 'es'
+              }
+            ]}
+            required
+            selectClasses="grid-cols-2 mb-4"
+          />
+
+          <TextInput
+            containerClasses="mb-4"
```
mdbiscan, line 185: `input type="email"` for emails.

client/src/_shared/forms/DropdownInput.js
```diff
@@ -0,0 +1,161 @@
...
+ * @param {func}     onChange            Callback to be triggered when the dropdown's selected option changes.
+ * @param {string}   [optionClasses]     Custom classes to be applied to the "options" in the dropdown.
+ * @param {Object[]} options             Array of options with a value (for direct comparison) and a label (for display).
+ * @param {string}   [placeholder]       Placeholder text to display inside the dropdown select box.
+ * @param {boolean}  [required]          Indicates whether or not the dropdowns's value is required.
+ * @param {string}   [selectClasses]     Custom classes to be applied to the "select" box div.
+ *
+ */
+
+export default function DropdownInput({
```
mdbiscan, line 24: An issue that custom drop downs have is that they suck on mobile devices. They're hard to use in the tiny screens, and they require extra work to make them less awful. If mobile devices are an important target, this is something to really consider.

client/src/Signup/Signup.js
```diff
@@ -0,0 +1,252 @@
...
+          <img
+            className="w-24 medium:w-48 mx-auto"
+            alt="Pie for Providers logo"
+            src={piefulltanlogo}
+          />
+        </div>
+        <h1 className="text-center my-8 medium:mt-16 large:text-left large:mt-12 large:mb-6">
+          <span className="font-bold underline">SIGN UP</span> or{' '}
+          <Link to="/login">LOG IN</Link>
+        </h1>
```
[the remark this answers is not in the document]
katelovescode: Yeah, I did it for accessibility reasons since the page had no header; this is the right solution, thank you!

client/src/_shared/forms/DropdownInput.js
katelovescode: Gonna make this a refactor ticket; right now we're not targeting mobile devices, but I'm curious about why/how they suck.  I know I'm on an iPhone, but I'd be curious what/how I should test so I can see the suckiness first hand for troubleshooting.

client/src/_shared/forms/SubmitButton.js
```diff
@@ -0,0 +1,38 @@
...
+  inputId,
+  label
+}) {
+  const containerClass = ['submit', containerClasses]
+    .filter(item => !!item)
+    .join(' ')
+
+  return (
+    <div className={containerClass}>
+      <button id={inputId} className={buttonClasses} type="submit">
```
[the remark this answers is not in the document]
katelovescode: Probably not!

client/src/_shared/forms/ToggleInput.js
```diff
@@ -0,0 +1,108 @@
...
+    .filter(item => !!item)
+    .join(' ')
+
+  const handleSelect = option => {
+    setSelectedOption(option.value)
+    onChange(option.value)
+  }
+
+  return (
+    <div className={containerClass}>
```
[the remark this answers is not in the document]
katelovescode: I will investigate this again - the dropdown/radio stuff I was reading indicated that you can't control mobile styling with the core <select> and radio type inputs (I may have not looked deep enough).  The alternative is to talk to my designer about allowing native mobile styling of those elements rather than needing them styled.  Will make a refactor ticket for this.

client/src/tailwind.css
```diff
@@ -0,0 +1,21 @@
+@tailwind base;
+
+@font-face {
```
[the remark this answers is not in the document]
katelovescode: Refactor ticket written; gonna have to get the .woff assets

katelovescode:
cc @mdbiscan 

Tailwind does do a reset by default that seems to work for us.  I set the base font-size. I prefer semantic-ish ordering rather than alpha, depending on the context.

I'm a little confused by this comment:

> I recommend not having separate class props on components. For simplicity and design adherence, it's a lot to interface with and I can see it getting abused. The components should only allow class props for positioning of the component in the content flow.

Do you mean you don't want to have styling classes on components?  I thought our original conversation was about allowing components to be reused and over-ridden with styles so we don't have to have, say, 3 or 4 different inputs where the only difference is the border, for example.  I might be misunderstanding the comment or the original conversation.

But all of this has been wonderful!  Thank you so much for all your advice and time. :)

PR #87, opened 2020-04-28
added get started page

nemiasalc56:
# 💅🏼 What issue does this fix?

# 🍂 Before You Submit:

 * [ ] Did you write tests?
 * [ ] Did you run `rails rswag` from the root?
 * [ ] Did you run `rubocop` from the root?
 * [ ] Did you run `yarn lint` in `/client`?
 * [ ] Did you run `yarn test` in `/client`?
 * [ ] Are your primary keys UUIDs on any new tables?

# 🛷 Deployment Considerations

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

# 🧵 Steps to set up locally

# 🧳 Steps to test

# 🕯 Caveats, concerns

Issue #65, opened 2020-04-22
As a user, I can see the "Get Started" page 


csprayregen:
## ✨ What new feature do you want to see?
Users can see the "Get Started" page

## 🧶 Acceptance Criteria

- [ ] -The design in the mockup is present
- [ ] I can click the "Get Started" button to be redirected to the next page
- [ ] The word "Amanda" in the mockup is replaced with the user's greeting_name field from the Users table

## 🥧 Additional context
Mockup: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=227%3A307
Tablet: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=313%3A1049
Mobile: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=619%3A2584

csprayregen:
Per KD, need to create separate ticket for the borders and layout of the page - Andrew created this already.

csprayregen:
- [ ]  The design in the mockup is present - not yet - issues noted below: 
Design questions:
- [ ] Across all screen sizes, the icons on the four cards are smaller than in the mockups 
- [ ] Across all screen sizes, the title text on all four cards should be bigger. In staging, I'm seeing 14px for title and body font. In mockups, I'm seeing 18px for title vs. 14 px. for body. I think those relative differences matter for usability. 
- [ ] Across all screen sizes, there is less left/right padding than in the mockups, but I believe this is because of the new layout grid. Could you confirm @rebeccakarasiktw ?
- [ ] In the Large screen size, the break points here are different than in the mockups, but I believe this to be because of the new layout grid. Could you just confirm this @rebeccakarasiktw ? (Did I use the word break point correctly!?)
- [ ] The title (Welcome to Pie...) has the same font size and weight as the mockups, but it looks bigger to me. Am I crazy?! Is this an issue related to typography stuff being handled elsewhere? @katelovescode @rebeccakarasiktw ?

Noting that I'm not doing visual QA on mobile, because we'll need to redirect users to a different screen here (#145)

- [ ]  I can click the "Get Started" button to be redirected to the next page
This doesn't do anything, but I think should be removed as an AC for this ticket, because we're going to work on connecting the screens later. Do you agree @katelovescode ?

- [ ]  The word "Amanda" in the mockup is replaced with the user's greeting_name field from the Users table
[...]

PR #148, opened 2020-07-18
WIP: Add GettingStarted page

cyranix:

References #65 

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test

1. Log in (or stub `pie-token`)
1. Navigate to `/getting-started`

## 🕯 Caveats, concerns

Supersedes #87 

Not related to primary objective, but `window.matchMedia` test shim has been applied globally via `setupTests.js`. The previous attempt to get this working has been removed, and one component-specific copy was also removed.


cyranix:
Current status: in a good place structurally, needing guidance on style implementation.

**Mobile:**  
Screen Shot 2020-07-18 at 1 07 01 PM

**Desktop:**  
Screen Shot 2020-07-18 at 1 07 09 PM

cyranix:
Latest progress...

**Mobile:**
Screen Shot 2020-07-18 at 3 09 13 PM

**Tablet:**
Screen Shot 2020-07-18 at 3 09 25 PM

**Desktop:**
Screen Shot 2020-07-18 at 3 09 45 PM

I haven't put quite as much effort into adjusting some of the spacing and font size. Mainly, I wanted to 1) reduce conflicts in editing `craco.config.js` and 2) see whether there would be a wrapper layout around this entire page, which might address some of the padding problems at the edge.

Definitely let me know if there's anything you'd like me to improve!

client/src/GettingStarted/GettingStarted.js
```diff
@@ -0,0 +1,109 @@
...
+import AssignmentIcon from '@material-ui/icons/Assignment'
+import BusinessIcon from '@material-ui/icons/Business'
+import CloudUploadIcon from '@material-ui/icons/CloudUpload'
+import PlaylistAddIcon from '@material-ui/icons/PlaylistAdd'
+
+export function GettingStarted() {
+  /*
+  TODO:
+  - layout/general
+    - ask about outer layout
```
katelovescode, line 13: You can start the outer component as a part of this pull request if you want to, but you can also push it off and write another ticket, up to you depending on how much scope you want to cover here.

client/src/setupTests.js
```diff
@@ -7,3 +7,20 @@ import Enzyme from 'enzyme'
 import Adapter from 'enzyme-adapter-react-16'
 
 Enzyme.configure({ adapter: new Adapter() })
+
+// window.matchMedia isn't implemented by JSDOM, but the responsive parts of
```
katelovescode, line 11: 😍

client/src/GettingStarted/GettingStarted.js
```diff
@@ -0,0 +1,109 @@
...
+    - change classnames to suit P4P
+    - double-check responsive breakpoints
```
katelovescode, line 15: If you look in `tailwind.config.js` we have our breakpoints laid out under "screens"

client/src/GettingStarted/GettingStarted.js
```diff
@@ -0,0 +1,109 @@
...
+  - content
+    - replace "Amanda" with user name
```
katelovescode, line 17: +1 and this leads to a maybe longer conversation about user data; right now we're storing user tokens in local storage and nothing else.  Do we want to store the user's `greeting_name` as well so that this page specifically doesn't have to make an api call?  I'm not super opinionated about frontend so I'd love some opinions on this.  cc @sassygrody @noragharris

client/src/GettingStarted/GettingStarted.js
cyranix: Ah, I think what I was confused about was which responsive system to use. This component is using the antd `Row` and `Col` approach, and the breakpoints are currently expressed like
```
<Col xs={24} sm={12} lg={6}>
```
On further review, it seems like a fair amount of existing code uses the Tailwind style like
```
<div className="grid grid-cols-1 medium:grid-cols-8 large:grid-cols-2">
```
Would it be preferable for me to switch systems?

client/src/GettingStarted/GettingStarted.js
cyranix: Logged this as a separate task, #154. 👍

cyranix:
I'll plan to rebase after #151 is merged, since there will be a small merge conflict. 

Please feel free to point out anything else that's not in accordance with project norms; I'd rather get on the same page as the rest of the crew than blindly do things "my way". 😁

client/src/GettingStarted/GettingStarted.js
katelovescode: Yeah, we're moving away from `Row` and `Col` because we want to use tailwind's layout classes instead.  Sorry I didn't catch that on first review!

cyranix:
Got it — I'll make those changes ASAP this week so we can get this over the
finish line.

katelovescode:
@all-contributors please add @cyranix for code

Issue #97, opened 2020-05-05
As a user, I see a screen that tells me to check my email after I create an account


csprayregen:
## ✨ What new feature do you want to see?
After I click "submit" on my account sign up page, I am redirected to a screen that tells me to check and verify my email. 

## 🧶 Acceptance Criteria

- [ ] The design in the mock up is present

## 🖍️ Implementation Details
This screen follows from the account sign up described in #54 
After seeing this screen, the user checks her email and goes through the process described in #93 

## 🥧 Additional context
Figma file: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=388%3A2485
Mobile mockup: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=588%3A2935
Tablet mockup:: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=388%3A2485

csprayregen:
@rebeccakarasiktw FYI, please link to design above when this exists.

csprayregen:
Issues here: 

- [ ] Even when I signed up with test@test.com, the message I receive assumes I signed up with chelsea@pieforproviders.com 

- [ ] It's not clear if anything is happening when I click to re-send the email

Will do a full QA on this page, but wanted to record while I'm here. Please hold off on responding. 
![Screen Shot 2020-08-14 at 12 58 33 PM](https://user-images.githubusercontent.com/26717304/90278907-d19f6400-de2d-11ea-89ea-1ccfadf734cb.png)

katelovescode:
cc @jontrainor couple of QA items in the design issues list!

@csprayregen for your functionality issues:

- Nothing happens when I click to re-send the email - how do I test this? Add entire flow w/ figma mock up: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=388%3A2485

This isn't implemented as a part of this ticket, and also isn't ticketed, so it'll need a new ticket, good catch.

 - Even when I signed up with test@test.com, the message I receive assumes I signed up with chelsea@pieforproviders.com'

When you say "the message I receive" do you mean "Is chelsea@pieforproviders.com your correct email"?  That's hardcoded in currently.  @jontrainor  we should pass the user's email from the form to the redirect, so we can display it correctly

- noreply email is not currently right - currently coming from tech@ so update either email or mock up

[...]

PR #150, opened 2020-07-18
#97 add signup confirmation page

jontrainor:

#97 - Added the signup confirmation page and added some override styles for Ant.

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test

route to `/confirmation` and 👀 the new content and there is a test to make sure it renders.

## 🕯 Caveats, concerns

image
image


client/src/Signup/Confirmation.js
```diff
@@ -1,9 +1,59 @@
...
+      <LabelImportantIcon
+        className="mr-1"
+        style={{ color: '#000', width: '16px' }}
+      />
+      {children}
     </div>
   )
 }
+
+ListItem.propTypes = {
```
sassygrody, line 24: For consistency with the other components, would you move this propTypes assignment to be after the function definition?

client/src/Signup/Confirmation.js
```diff
@@ -1,9 +1,59 @@
...
+      <Title>Thanks for signing up!</Title>
+      <Title level={3}>We’ve sent you an email to verify your account.</Title>
+      <Divider />
+      <div className="text-left">
+        <div className="mb-2">
+          <Text>{"Didn't receive the email?"}</Text>
+        </div>
+        <ListItem>
+          <Text>
+            {`Is ${userEmail} your correct email without typos? If not, you can restart the signup process.`}
```
sassygrody, line 40: JSX interpolation doesn't require the {} around the string, so it could just be 

```js
 <Text>
   Is {userEmail} your correct email without typos? If not, you can restart the signup process.
</Text>
```

client/src/Signup/Confirmation.js
```diff
@@ -1,9 +1,60 @@
...
+
+export default function Confirmation() {
+  return (
+    <>
+      <Title>Thanks for signing up!</Title>
+      <Title level={3}>We’ve sent you an email to verify your account.</Title>
+      <Divider />
+      <div className="text-left">
+        <div className="mb-2">
+          <Text>{"Didn't receive the email?"}</Text>
```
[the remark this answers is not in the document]
jontrainor: I went to remove that and remembered why I did that. The linter doesn't like the `'`.

```
/Users/jontrainor/projects/pieforproviders/client/src/Signup/Confirmation.js
  31:21  error  `'` can be escaped with `&apos;`, `&lsquo;`, `&#39;`, `&rsquo;`  react/no-unescaped-entities

✖ 1 problem (1 error, 0 warnings)

```

so I thought this was more clear than using an escaped string like that. If you'd prefer the escaped string I can fix it though!

client/src/Signup/Confirmation.js
```diff
@@ -1,9 +1,57 @@
 import React from 'react'
+import PropTypes from 'prop-types'
+import { Divider, Typography } from 'antd'
+import LabelImportantIcon from '@material-ui/icons/LabelImportant'
 
-export default function Confirmation() {
+const { Title, Text, Link } = Typography
+
+const userEmail = 'chelsea@pieforproviders.com'
```
noragharris, line 8: Thoughts on moving these to heroku config variables? @katelovescode

client/src/Signup/Confirmation.js
```diff
@@ -1,9 +1,57 @@
...
+      <LabelImportantIcon
+        className="mr-1"
+        style={{ color: '#000', width: '16px' }}
+      />
+      {children}
     </div>
   )
 }
+
+export default function Confirmation() {
```
noragharris, line 23: looking at the code we've been following component arrow functions as opposed to functions!

client/src/Signup/Confirmation.js
jontrainor: I could totally see that for the `pieEmail`. I was thinking 'userEmail` is just a placeholder for the actual user that had signed up's email. I'll add a comment to make that clear!

client/src/Signup/Confirmation.js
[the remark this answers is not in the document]
katelovescode: No, given the choice I'd prefer what you have already 😂

katelovescode:
@all-contributors please add @jontrainor for code, doc, and tests

Issue #153, opened 2020-07-22
Convert a test to use react-testing-library instead of enzyme for the team to see if they like it better


jontrainor:
## 📋 Summary
Convert a test to use react-testing-library so the team can see if they like it better than enzyme.

## 💻 Implementation Details
https://testing-library.com/docs/react-testing-library/intro

## 🧐 Why tho
In my experience, RTL has been easier to use and read than enzyme for react tests.

katelovescode:
@jontrainor I think this can be marked approved, right?

PR #155, opened 2020-07-26
#153 convert enzyme tests to use react-testing-library for team to evaluate

jontrainor:
#153 
I chatted with @katelovescode about using RTL instead of enzyme and she was open to seeing a PR to evaluate. I made this change about a year ago at my day job and i have found it has been much easier to work. The main difference to me is it has one `render` method instead of enzyme's `shallow`, `mount`, and `render`, which each have their own quirks and difficulties. The docs are pretty great too https://testing-library.com/docs/react-testing-library/cheatsheet

Another thing I like about RTL is it encourages the developer to test the actual output the user will see vs the components the developer expects to be there or functions that were expected to run. One of my colleagues described it to me as only testing what a user will see, which I find to be a really helpful guide for what to test. So instead of testing that a div with a specific class exists I changed the tests to look for text on the page. Often times if there is no text specific to the thing I'm trying to test, I have found that can indicate a potential accessibility issue for people that rely on screen readers. I added some screen reader only text (with the tailwind class `sr-only`) in the `CSVImportReview` page to help test and improve accessibility (double win 🎉 !!)

## 🍂 Before You Submit

* [x ] Did you write tests?
* [x ] Did you run `yarn lint` in `/client`?
* [ x] Did you run `yarn test` in `/client`?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally
run `yarn` on `./client` to remove `enzdependenciesncies and install `rtl`

## 🧳 Steps to test

run tests

## 🕯 Caveats, concerns

I was originally going to just convert one but it was quick enough to sweep it at once so I thought that might help folks evaluate better. If the consensus is we would rather stick with `enzyme` then it'll be just as easy to toss out =)

### DIFF
--- a/client/src/CSVImport/CSVImportReview.js
+++ b/client/src/CSVImport/CSVImportReview.js
@@ -5,8 +5,10 @@ import PropTypes from 'prop-types'
 export function CSVImportReview({ kids }) {
   return (
     <div className="csv-import-review">
+      <h1 className="sr-only">Review Imported CSV</h1>
       {kids && (
         <div className="kids-list">
+          <span className="sr-only">List of kids</span>
           {kids.map(kid => {
             return (
               <div key={sha1(kid[0], kid[1], kid[2])}>
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Learn more at [www.pieforproviders.com](http://www.pieforproviders.com)
   - API Documentation with swagger
 - Frontend: React
   - ESLint/Prettier
-  - Jest/Enzyme
+  - Jest/React Testing Library
   - Husky for pre-commit hooks
   </details>
 
--- a/client/package.json
+++ b/client/package.json
@@ -69,9 +69,6 @@
     }
   },
   "jest": {
-    "snapshotSerializers": [
-      "enzyme-to-json/serializer"
-    ],
     "collectCoverageFrom": [
       "**/*.{js,jsx}",
       "!**/node_modules/**",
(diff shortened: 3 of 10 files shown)


client/src/CSVImport/CSVImportReview.js
```diff
@@ -5,8 +5,10 @@ import PropTypes from 'prop-types'
 export function CSVImportReview({ kids }) {
   return (
     <div className="csv-import-review">
+      <h1 className="sr-only">Review Imported CSV</h1>
```
katelovescode, line 8: 😮

weedySeaDragon:
Just a little nerdy software engineering comment about testing:

>  One of my colleagues described it to me as only testing what a user will see, which I find to be a really helpful guide for what to test. So instead of testing that a div with a specific class exists I changed the tests to look for text on the page.

Testing only what a user will see/interact with is generally called **"black box"** testing (you don't deal with the internals).  This is usually done when testing a specific _user story_, user journey, _use case_, and/or doing _acceptance testing._

It can be helpful to organize your actual testing code along those lines so it's clear what should and should not be tested.  (Ex:  if you're doing a _user story_ test, you should be testing only what the user experiences (sees, interacts with, etc.) and not the specific elements that are on the page that support/accomplish that (`divs`, `buttons` etc.)
Note that _accessibility testing_ can definitely be done as _black box_ testing; you are testing how the system is presented (e.g. how a screen reader will render the page) and the interactions with it (e.g. tabbing through a page).
There are sometimes some _white box_ tests that are written to ensure that particular elements exist that need to be there for universal accessibility.  Ex:  ensure there are alt text tags, elements needed for coherent text-use of a page (usually tabbing), semantic hierarchies are coherent, etc.

Perhaps you all already know all of that.   I just couldn't stop myself from "sharing" :-D

Issue #168, opened 2020-08-02
Parent abstract class for ActiveRecord models with UUID 'created_on' 


weedySeaDragon:
## 📋 Summary

All of the ActiveRecord models that are using UUID need a line to 'fix' the `.first` and `.last` behavior.
Instead of repeating the code in all of those classes, abstract it out. 

## 💻 Implementation Details

1. Put
```
  # Handles UUIDs breaking ActiveRecord's usual ".first" and ".last" behavior
  self.implicit_order_column = 'created_at'
```
into a simple abstract parent class 
2. Change existing ActiveRecord models to inherit from that class.

## 🧐 Why tho

 DRY: write it once, share it by many, etc etc.

PR #169, opened 2020-08-02
#168: Add an abstract class for all ActiveRecords using UUID so ordering is correct for all

weedySeaDragon:

#168 Parent abstract class for ActiveRecord models with UUID 'created_on' 
1. Added `class UuidApplicationRecord < ApplicationRecord` 
   It has the common bit of code that all AR models using UUID need so that ordering (.first, .last) works properly.
2. Changed existing ActiveRecord models to inherit from this new class and removed the duplicated code.

## 🍂 Before You Submit

* [ ] Did you write tests? n/a
* [ ] Did you run `rails rswag` from the root? n/a
* [ ] Did you run `rubocop` from the root? n/a
* [ ] Did you run `yarn lint` in `/client`? n/a
* [ ] Did you run `yarn test` in `/client`? n/a
* [ ] Are your primary keys UUIDs on any new tables? n/a

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally
_(none)_

## 🕯 Caveats, concerns

All new AR models (with UUIDs) will need to inherit from this class instead of ApplicationRecord.  [just a reminder]

I'm happy to use a different name for the class.  This is just made sense to me.

### DIFF
--- a/app/models/business.rb
+++ b/app/models/business.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # The businesses for which users are responsible for keeping subsidy data
-class Business < ApplicationRecord
+class Business < UuidApplicationRecord
   CATEGORIES = %w[
     licensed_center_single
     licensed_center_multi
@@ -11,9 +11,6 @@ class Business < ApplicationRecord
     license_exempt_center_single
     license_exempt_center_multi
   ].freeze
-
-  # Handles UUIDs breaking ActiveRecord's usual ".first" and ".last" behavior
-  self.implicit_order_column = 'created_at'
 
   belongs_to :user
 
--- a/app/models/child.rb
+++ b/app/models/child.rb
@@ -2,7 +2,6 @@
 
 # A child in care at businesses who need subsidy assistance
 class Child < UuidApplicationRecord
-
   belongs_to :user
 
   validates :active, inclusion: { in: [true, false] }
--- a/app/models/child.rb
+++ b/app/models/child.rb
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
 
 # A child in care at businesses who need subsidy assistance
-class Child < ApplicationRecord
-  # Handles UUIDs breaking ActiveRecord's usual ".first" and ".last" behavior
-  self.implicit_order_column = 'created_at'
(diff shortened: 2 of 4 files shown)


katelovescode:
This looks good - are there any tests we could write to verify which of the results come first when using UuidApplicationRecord?

weedySeaDragon:
Yup, definitely.  But since I explicitly declared this as an abstract class, we can't instantiate it.
How to test is a a bit of a judgement call:
1. We can test by using a (concrete) class we _know_ will be a subclass, and we can put the test in a file named  `uuid_active_record_spec.rb` so it's clear that we're testing _that_ behavior.
2. We can test by using a (concrete) class we _know_ will be a subclass,  and put the test in a file named for the concrete class to make it explicit that we are hardwiring this test to this concrete class
_or_
3. I can remove the `self.abstract_class = true` statement from UuidActiveRecord and then we can go ahead and write test for that class directly.

Preferences?  Thoughts?
(I would go with (3), but that's my design/coding preference. I will be happy to implement whatever.  :-)  )

katelovescode:
> 1. We can test by using a (concrete) class we _know_ will be a subclass, and we can put the test in a file named  `uuid_active_record_spec.rb` so it's clear that we're testing _that_ behavior.

I'm actually leaning toward option 1. I found a couple of solutions here: https://stackoverflow.com/questions/28114647/testing-abstract-classes-in-rspec/28114785  Is this something we could do?
```
let(:AnonymousUuidBackedClass) do
    Class.new(UuidApplicationiRecord) do
      # ...
    end
  end
```

You don't have to

weedySeaDragon:
Thinking about this more, I don't think we should do a test for this.
because we would effectively be testing the `ActiveRecord` `implicit_order_column` method.  
Which is not our responsibility. (I trust it's been tested and works. :-) )

weedySeaDragon:
I don't mean to be or sound argumentative. I'll definitely do whatever you'd like. :-)

katelovescode:
> I don't mean to be or sound argumentative. I'll definitely do whatever you'd like. :-)

Really good point, actually, not at all argumentative!  I'll approve and merge!

Issue #162, opened 2020-08-02
DB Implementation: Agencies


katelovescode:
## 📋 Summary
Write the migration and model for Agencies from the data model.

## 💻 Implementation Details
- This will have no route; these records will be manually added to the db either through rake tasks or SQL
- Only admins with direct access to the db will be able to update these records
- see docs/dbdiagram.dbml for field definitions and associations

## 🧐 Why tho
Start getting the DB built to make front end tickets easier

weedySeaDragon:
@katelovescode : Since these will only be updated externally (rake, direct to db, etc), 
then are these _read only_ ?  In which case the only controller actions should be `index` and `show` ?

weedySeaDragon:
@katelovescode also:  Does the `state` attribute refer to a State in the United States like "Illinois" ?   
or does it refer to a "status" ?
(So I know what to put in `db/seeds.rb` and `spec/factories/agencies.rb`)

PR #170, opened 2020-08-02
#162 DB Implementation: Agencies

weedySeaDragon:

#162 Add `Agencies` to the db model

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally
_(none)_

### Open questions/Notes:
Since `agencies` data will be added externally (e.g. rake tasks, direct db access, etc.),
and since there should be routes for them:
1.  Which controller actions are needed?  any of them?
2.  I didn't add a `spec/requestes/api/v1/agencies_spec.rb` file (since there aren't any routes for them).
3. I did add the `.json` schemas so that other models will be able to reference them
4. Should I also add a definition for `agency` in `spec/swagger_helper.rb` so that other models can refer to one in the future?  ex:
```
 definitions: {
        agency: { 
          type: :object,
          properties: {
            name: { type: :string, example: 'Agency 1' },
            state: { type: :string, example: 'IL'},
            active: { type: :boolean, example: 'true' }
          }
        },
```


app/controllers/api/v1/agencies_controller.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+# API for agencies.  Note that this class is 'read only' for client apps.
+#   Data required for Agencies will be added to the db via rake tasks or other external means.
+class Api::V1::AgenciesController < Api::V1::ApplicationController
```
katelovescode, line 5: I think for now we should remove the controller actions because we'll be using rake tasks.  If we need routes later on, I'd rather add them in then.  I'd love to see a little documentation added to this effect; could we start a `docs/????.md` file describing caveats and architectural decisions?  I'm not 100% sure what to call it 🤔

app/models/agency.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+# Agencies provide child care subsidy payments
+class Agency < ApplicationRecord
+  # Handles UUIDs breaking ActiveRecord's usual ".first" and ".last" behavior
+  self.implicit_order_column = 'created_at'
```
katelovescode, line 6: Now that #169 is merged in we should inherit from `UuidApplicationRecord`

db/seeds.rb
```diff
@@ -22,4 +22,13 @@
   name: "Happy Seedlings Childcare",
   category: "licensed_center_single",
   user: user
-)
\ No newline at end of file
+)
+
+agency_1 = Agency.create!(
+    name: "Agency 1",
```
katelovescode, line 28:
```suggestion
    name: "Community Child Care Connection",
```

db/seeds.rb
```diff
@@ -22,4 +22,13 @@
...
+    state: "IL"
+)
+agency_2 = Agency.create!(
+    name: "Agency 2",
```
katelovescode, line 32:
```suggestion
    name: "Children's Aid and Family Services",
```

db/seeds.rb
```diff
@@ -22,4 +22,13 @@
...
+    state: "IL"
```
katelovescode, line 33:
```suggestion
    state: "MA"
```

spec/factories/agencies.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :agency do
+    name { 'Agency name' }
```
katelovescode, line 5: I'd love to see a couple of semantic options here, see `spec/factories/businesses.rb#5` - those semantic options go in en.yml under the Faker name key, with their own custom key.

Here's a list of suggestions:
```
Children's Aid and Family Services
Community Child Care Connection
Maui Family Support Services
Wisconsin Shares
Child Care Aware of America
```
```suggestion
    name { Faker::Name.agencies }
```

spec/factories/agencies.rb
```diff
@@ -0,0 +1,21 @@
...
+    state { 'IL' }
```
katelovescode, line 6:
```suggestion
    state { Faker::Address.state_abbr }
```

app/models/agency.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  validates :active, inclusion: { in: [true, false] }
+  validates :name, presence: true
```
katelovescode, line 9: State is also required!

```suggestion
  validates :name, presence: true
  validates :state, presence: true
```

db/migrate/20200802210346_create_agencies.rb
```diff
@@ -0,0 +1,11 @@
+class CreateAgencies < ActiveRecord::Migration[6.0]
+  def change
+    create_table :agencies, id: :uuid do |t|
+      t.string :name, null: false
+      t.string :state
```
katelovescode, line 5:
```suggestion
      t.string :state, null: false
```

spec/models/agency_spec.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Agency, type: :model do
+  it { should validate_presence_of(:name) }
```
katelovescode, line 6:
```suggestion
  it { should validate_presence_of(:name) }
  it { should validate_presence_of(:state) }
```

spec/support/api/schemas/agencies.json
```diff
@@ -0,0 +1,21 @@
...
+  "type": "array",
+  "required": ["agencies"],
+  "items": {
+    "type": "object",
+    "required": [
+      "active",
+      "created_at",
+      "id",
+      "name",
+      "updated_at"
```
katelovescode, line 11:
```suggestion
      "state",
      "updated_at"
```

spec/support/api/schemas/agencies.json
```diff
@@ -0,0 +1,21 @@
...
+    ],
+    "properties": {
+      "active": { "type": "boolean" },
+      "created_at": { "type": "string" , "format": "date-time" },
+      "id": { "type": "string" },
+      "name": { "type": "string" },
```
katelovescode, line 17:
```suggestion
      "name": { "type": "string" },
      "state": { "type": "string" },
```

spec/support/api/schemas/agency.json
```diff
@@ -0,0 +1,16 @@
+{
+  "type": "object",
+  "required": [
+    "created_at",
+    "id",
+    "name",
+    "updated_at"
```
katelovescode, line 7:
```suggestion
    "state",
    "updated_at"
```

spec/support/api/schemas/agency.json
```diff
@@ -0,0 +1,16 @@
...
+  ],
+  "properties": {
+    "active": { "type": "boolean" },
+    "created_at": { "type": "string" , "format": "date-time" },
+    "id": { "type": "string" },
+    "name": { "type": "string" },
```
katelovescode, line 13:
```suggestion
    "name": { "type": "string" },
    "state": { "type": "string" },
```

db/migrate/20200802210346_create_agencies.rb
```diff
@@ -0,0 +1,11 @@
...
+      t.boolean :active, null: false, default: true
+
+      t.timestamps
+    end
+  end
```
katelovescode, line 10: Can we add a compound unique index here so no one can add an agency twice?

Something like
```
add_index :agencies, [:name, :state], unique: true
```

katelovescode:
> 1. Which controller actions are needed?  any of them?
I don't think so, as a part of this PR

> 2. I didn't add a `spec/requestes/api/v1/agencies_spec.rb` file (since there aren't any routes for them).
👍 

> 3. I did add the `.json` schemas so that other models will be able to reference them
👍 

> 4. Should I also add a definition for `agency` in `spec/swagger_helper.rb` so that other models can refer to one in the future?
I don't think so.  Swagger is specifically to document the API and for now this model is not available via API

app/controllers/api/v1/agencies_controller.rb
weedySeaDragon: I'll create one:  technical-decisions.md  (we can always rename it later).

I'd suggest we keep it simple for now and then later you can decide if it needs to be more structured / formalized, etc.
(There are, in fact, some defined structures to us for these. But that's waaay overkill. :-)  )
My experience has been that knowing where to do the search(es) to find an answer where you wonder "why did we do it like this?!?"  is 80% of what you want. (or even 90%).  YMMV

weedySeaDragon:
For whatever reason, GitHub says you need to review and approve again:
`weedySeaDragon dismissed katelovescode’s stale review via 88ffd18 5 minutes ago`

Not sure why my commit caused this to happen.  I don't think I clicked on anything explicitly; I just merged and pushed a new commit.

Issue #16, opened 2020-02-22
Money Handling


katelovescode:
## 📋 Summary
Spike on how we're going to handle money calculation

## 💻 Implementation Details
investigate https://github.com/RubyMoney/money-rails

## 🧐 Why tho
Rounding in money is complex.  Libraries do this work for us already.

## 📎 Deliverables
- choose a money handling gem
- write tickets to implement that library

Issue #165, opened 2020-08-02
DB Implementation: Payments


katelovescode:
## 📋 Summary
Write the migration, model and controller actions necessary for basic CRUD on Payments from the data model.

## 💻 Implementation Details
- This is an authenticated route
- Users will be managing payments on their account and will need authorization (i.e. they can only add/edit/update/delete payments associated to their account)
- All CRUD actions are required
- see docs/dbdiagram.dbml for field definitions and associations
- technically this work is blocked by #162 and #159 but the association can probably be made as a part of those tickets if this work comes first; if you pick up this ticket and 159 or 162 has no merged PR, please update the appropriate tickets to ensure the association is made

## 🧐 Why tho
Start getting the DB built to make front end tickets easier

weedySeaDragon:
@katelovescode  - just wondering -- should this be closed now?  Just learning what the overall workflow is.  (And I assume you should close it, yes?)

PR #171, opened 2020-08-03
#165 DB Implementation: Payments

weedySeaDragon:

#165 DB Implementation: Payments

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🕯 Caveats, concerns

1. I rebased this onto the _features/159/add_sites_and_child_sites_  branch and cherry picked commits from the _features/162/add_agencies_ branch  so that I'd have Sites and Agencies to work with.  Once those are merged into `develop` and this PR is whipped into shape I'll fix it (I'll rebase this properly onto `develop`).

1. I wasn't sure how you wanted to handle _money_.  ("monetize" items in the db diagram.)  I took a leap and installed the **`money-rails` gem.**

2. Are there other attributes for a Payment that must be non-null?  (e.g. in the migration + validations)

2. How should the `slug` for a Payment be constructed?  (How to guarantee uniqueness?)  Ex:
   - what if 2 partial payments are made from an agency to cover a period of care at a site, and those payments came in on the same day?   [I know so little about the domain, so I'm WAGging here.]
 (See the `TODO` in `app/models/payment.rb`)

4. I used _references_ in `spec/swagger_helper.rb`.  Is that ok?  

I wanted to go ahead and submit this so that I can also understand if I've done all of the swagger stuff correctly.


spec/factories/payments.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :payment do
+    paid_on { Faker::Date.between(from: care_finished_on.strftime('%Y-%m-%d'), to: "#{Time.zone.today.strftime('%Y-%m-%d')}") }
+    care_started_on { Faker::Date.backward(365).strftime('%Y-%m-%d') }
+    care_finished_on { Faker::Date.between(from: care_started_on.strftime('%Y-%m-%d'), to: "#{Time.zone.today.strftime('%Y-%m-%d')}").strftime('%Y-%m-%d') }
+    amount { Faker::Number.between(from: 0, to: 99_999_999) }
+    discrepancy { Faker::Number.between(from: 0, to: 999_999) }
```
katelovescode, line 9: For accuracy's sake, I'd love to save the amount to a variable and then get a discrepancy that's less than that amount.

i.e.

```
FactoryBot.define do
  payment_amount = Faker::Number.between(from: 1, to: 99_999_999)
  factory :payment do
    ...
    amount { payment_amount }
    discrepancy { Faker::Number.between(from: 0, to: payment_amount) }
```

db/migrate/20200802222331_create_payments.rb
```diff
@@ -0,0 +1,18 @@
+class CreatePayments < ActiveRecord::Migration[6.0]
+  def change
+    create_table :payments, id: :uuid do |t|
+      t.date :paid_on
+      t.date :care_started_on
+      t.date :care_finished_on
+      t.monetize :amount
```
katelovescode, line 7: I don't know that we'll ever want to accept a payment of less than 1 cent, so we should put that validation in here for the amount being greater than 0?  I'm not sure if monetize can do that, but if we could do it at the db level and on the model, that'd be amazing.

spec/models/payment_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Payment, type: :model do
+  it { should belong_to(:site) }
+  it { should belong_to(:agency) }
+
```
katelovescode, line 8: Need the not null validations, and if possible, the greater-than-zero validation

spec/support/api/schemas/payment.json
```diff
@@ -0,0 +1,31 @@
+{
+  "type": "object",
+  "required": [
+    "agency_id",
+    "amount_cents",
```
katelovescode, line 5: are the `amount_currency` and `discrepancy_currency` also technically required?

spec/requests/api/v1/payments_spec.rb
```diff
@@ -0,0 +1,316 @@
...
+      "county": 'Harrison',
+      "qris_rating": '4',
+      "business_id": business_id
+    }
+  end
+  let(:site_id) { Site.create!(site_params).id }
+  let!(:payment_params) do
+    {
+      "agency_id": agency_id,
+      "amount_cents": '123400',
```
katelovescode, line 53: Couple questions:

Do the `amount_cents` and `discrepancy_cents` fields come in as parameters or should it come in as `amount` when hitting the api?

If it should be `amount_cents` and `discrepancy_cents`, do we also need to pass `amount_currency` and `discrepancy_currency`?

spec/swagger_helper.rb
```diff
@@ -33,6 +33,21 @@
       #   }
       # },
       definitions: {
+        payment: {
+          type: :object,
+          properties: {
+            agency_id: { type: :string, example: '3fa57706-f5bb-4d40-9350-85871f698d52' },
+            amount_cents: { type: :integer, example: '19999' },
```
katelovescode, line 40: Same question as for above in the spec params; should this really take in `amount_cents` or just `amount`?

katelovescode:
> 1. I rebased this onto the _features/159/add_sites_and_child_sites_  branch and cherry picked commits from the _features/162/add_agencies_ branch  so that I'd have Sites and Agencies to work with.  Once those are merged into `develop` and this PR is whipped into shape I'll fix it (I'll rebase this properly onto `develop`).

👍 

> 2. I wasn't sure how you wanted to handle _money_.  ("monetize" items in the db diagram.)  I took a leap and installed the **`money-rails` gem.**

Perfect!  That means this PR will close #16 as well 😂 

> 3. Are there other attributes for a Payment that must be non-null?  (e.g. in the migration + validations)

In `docs/dbdiagram.dbml` I've put the expected field constraints for every field.  Here's the layout for payments:

```
Table payments {
  id uuid [pk, unique, not null]
  agency_id uuid [ref: > agencies.id, not null]
  site_id uuid [ref: > sites.id, not null]
  paid_on date [not null]
  care_started_on date [not null]
  care_finished_on date [not null]
  amount monetize [not null]
  slug string [not null]
  discrepancy monetize
}
```

> 4. How should the `slug` for a Payment be constructed?  (How to guarantee uniqueness?)

Personally I think the best way to do it is agency_id, site_id, paid_on, amount.  The likelihood of that being non-unique is very low.  You can add a random string to the slug method as well if you think it will add more confidence of uniqueness, as well.

> 5. I used _references_ in `spec/swagger_helper.rb`.  Is that ok?
I don't see references in swagger_helper, am I missing something?

spec/requests/api/v1/payments_spec.rb
weedySeaDragon: Hm.  Good questions.  I'm not sure; will investigate.

spec/factories/payments.rb
weedySeaDragon: Don't have to even save it to a variable.  You can define amounts based on other attributes.  
ex: `discrepancy { Faker::Number.between(from: -0, to: amount) }`
(which is how I have it defined ATM.)

spec/requests/api/v1/payments_spec.rb
weedySeaDragon: I checked -- only need to use the attribute name.  Only the `money-rails` gem needs to use the `_cents` and `_currency` fields.  (Unless you're doing something specific.  We're not.)

I'll update accordingly.

spec/support/api/schemas/payment.json
weedySeaDragon: nope.  removed

spec/requests/api/v1/payments_spec.rb
weedySeaDragon: OK ...  you _do_ need to specify the `_cents`   (Makes more sense that you do.  Then it's very clear that the `integer` that is being sent to/from literally is an amount is _cents._)

spec/swagger_helper.rb
weedySeaDragon: ..._cents

app/controllers/api/v1/payments_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    @payment = Payment.find_by!(slug: params[:slug])
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  # Only allow a trusted parameter "white list" through.
+  def payment_params
+    params.require(:payment).permit(:paid_on,
+                                    :care_started_on,
+                                    :care_finished_on,
+                                    :amount,
```
katelovescode, line 57: Not to be nitpicky, but if we have to send `..._cents`, will `amount` or `discrepancy` ever be sent as params?

app/models/payment.rb
```diff
@@ -0,0 +1,45 @@
...
+  belongs_to :agency
+  belongs_to :site
+
+  validates :amount, numericality: { greater_than: 0.00 }
+  validates :care_finished_on, presence: true # TODO: use Date validator
+  validates :care_started_on, presence: true # TODO: use Date validator
+  validates :discrepancy, numericality: true
+  validates :paid_on, presence: true # TODO: use Date validator
+
+  before_validation { |payment| payment.slug = generate_slug("#{payment.care_started_on}#{payment.site_id}#{payment.agency_id}") }
```
katelovescode, line 14: Can we change this slug to `payment.site_id`, `payment.paid_on`, `payment.agency_id` and `payment.agency_amount` per comments from before?

db/seeds.rb
```diff
@@ -18,8 +18,94 @@
...
-business = Business.create!(
+business_happy_seedlings = Business.create!(
   name: "Happy Seedlings Childcare",
   category: "licensed_center_single",
   user: user
-)
\ No newline at end of file
+)
+
+site_happy_seeds_little_oaks = Site.create!(
```
katelovescode, line 27: Check the seeds I've got in the #175 PR; I've shifted from `create!` and have a few sites seeded.

spec/requests/api/v1/payments_spec.rb
```diff
@@ -0,0 +1,315 @@
+# frozen_string_literal: true
+
+require 'swagger_helper'
+
+RSpec.describe 'payments API', type: :request do
+  let!(:user_params) do
```
katelovescode, line 6: I did a tiny refactor of the associated model records in #175 if you want to take a look and make this less verbose :)

spec/swagger_helper.rb
```diff
@@ -32,7 +32,22 @@
       #     scopes: {}
       #   }
       # },
+      # rubocop:disable Style/NumericLiterals
       definitions: {
+        payment: {
+          type: :object,
+          properties: {
+            agency_id: { type: :string, example: '3fa57706-f5bb-4d40-9350-85871f698d52' },
+            amount: { type: :integer, minimum: 0, example: 19999 },
```
katelovescode, line 41: Rather than disabling rubocop can we use the underscore style for the larger numbers?

```suggestion
            amount: { type: :integer, minimum: 0, example: 19_999 },
```

app/controllers/api/v1/payments_controller.rb
weedySeaDragon: I thought about this, too.  (This isn't being nitpicky. At least not in my book.)
It is legitimate to interact with the attribute without the `_cents` or `_currency` suffixes, e.g. in specs or non-strict API, as in a rake task.

However, we can certainly restrict interaction to only the `_cents` and `_currency` -suffixed parameters for now.
It's your call.  Either way is fine.  And I can make a note in a block comment for this method so it's clear either way.

app/models/payment.rb
weedySeaDragon: oops. missed that.  will do.

app/models/payment.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  # The money-rails gem specifically requires that the '_cents' suffix be
+  # specified when using the "monetize" macro even though the attributes are
+  # referred to without the '_cents' suffix.
+  # IOW, you only need to refer to payment.amount or payment.discrepancy ,
+  # unlike the following statements.
```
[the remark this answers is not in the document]
weedySeaDragon: That's part of working with the `money` gem.  I thought it was worth noting so that future readers would be clued in to that.  It's definitely something that they should fix with the gem so they're consistent everywhere.  This is often a source of confusion.

db/seeds.rb
weedySeaDragon: Will do.

spec/swagger_helper.rb
weedySeaDragon: We definitely can.  
I'm not a fan of this style (hence my resistance). :-)
(It's also representing 'One-hundred, nineteen dollars and ninety-nine cents'  in my head.)

However:  this is your project and you get to set the styles.  :-)  
So I'll make the change.

db/seeds.rb
weedySeaDragon: Just curious:  why not the `!` version of `first_or_create`?  Why not use `first_or_create!` so you explicitly get an error if the create fails?

app/controllers/api/v1/payments_controller.rb
katelovescode: Ahhh, you know what, you make a good point, I wasn't thinking about raw access to the controller not via the API.  I'm good with leaving them both in; maybe add in the _currency too while we're add it?  Never know if we'll want to change it........ 🤔

app/models/payment.rb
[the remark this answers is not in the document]
katelovescode: My gif reaction was to the money gem itself, not to you or your comments (which are wonderful!) 😂

db/seeds.rb
katelovescode: Mostly because seeding is a non-typical action, but that's a valid point, do you mind tweaking them across the whole seed file?

spec/swagger_helper.rb
katelovescode: Definitely get it, thanks for being flexible! :)

spec/swagger_helper.rb
```diff
@@ -200,7 +213,9 @@
             }
           }
         }
-      }
+      },
+      createPayment: { '$ref': '#/definitions/payment' },
```
katelovescode, line 217: One quick note; all those fields shouldn't be required on update; wonder if there's a way to DRY this up and only apply "required" to the create action 🤔

katelovescode:
Will review again after merge conflicts are fixed so we don't have to do it twice

spec/swagger_helper.rb
weedySeaDragon: this is where my Swagger knowledge is lacking.
Does that mean that only the 
```
 amount: { type: :integer, minimum: 0, example: 19_999 },
            care_finished_on: { type: :string, example: '2020-03-15' },
            care_started_on: { type: :string, example: '2020-01-01' },
            discrepancy: { type: :integer, example: 456 },
            paid_on: { type: :string, example: '2020-05-20' },
```
should be in the list of fields for an update?

You can refer to something you've previously defined with a  `'$ref': '...'` statement.   It just does a literal substitution (plops whatever the thing is right there at your `'$ref'...` statement.
Perhaps you knew that already. :-)

spec/swagger_helper.rb
```diff
@@ -37,14 +37,16 @@
...
             care_finished_on: { type: :string, example: '2020-03-15' },
             care_started_on: { type: :string, example: '2020-01-01' },
-            discrepancy: { type: :integer, example: 456 },
+            discrepancy_cents: { type: :integer, example: 456 },
+            discrepancy_currency: { type: :string,  example: 'USD' },
             paid_on: { type: :string, example: '2020-05-20' },
             site_id: { type: :string, example: '3fa57706-f5bb-4d40-9350-85871f698d51' }
           },
-          required: %w[agency_id site_id amount care_finished_on care_started_on discrepancy paid_on]
+          required: %w[agency_id site_id amount_cents care_finished_on care_started_on discrepancy_cents paid_on]
```
katelovescode, line 49: `discrepancy_cents` should be optional on this one

app/models/payment.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+# Payments made by an agency to a provide for a particular site
+class Payment < UuidApplicationRecord
+  belongs_to :agency
+  belongs_to :site
+
+  validates :amount, numericality: { greater_than: 0.00 }
+  validates :care_finished_on, date_param: true
```
katelovescode, line 9: For these fields can we also get presence validators? (amount, care_finished_on, care_started_on, paid_on)

app/controllers/api/v1/agencies_controller.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+# API for agencies.  Note that this class is 'read only' for client apps.
+#   Data required for Agencies will be added to the db via rake tasks or other external means.
```
katelovescode, line 4: Make sure you remove this file since we decided in #170 not to have controller actions

db/migrate/20200802222331_create_payments.rb
```diff
@@ -0,0 +1,18 @@
+class CreatePayments < ActiveRecord::Migration[6.0]
+  def change
+    create_table :payments, id: :uuid do |t|
+      t.date :paid_on, null: false
+      t.date :care_started_on, null: false
+      t.date :care_finished_on, null: false
+      t.monetize :amount, null: false, default: 0
```
katelovescode, line 7: Can we remove the default here so if it's blank it won't save?

spec/models/payment_spec.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Payment, type: :model do
+  it { should belong_to(:site) }
+  it { should belong_to(:agency) }
+  it { should validate_numericality_of(:amount).is_greater_than(0.00) }
+  it { should validate_numericality_of(:discrepancy) }
```
katelovescode, line 9: If we add the presence validations in the model, make sure to add the specs for that here

spec/swagger_helper.rb
katelovescode: Ya!  So, look at the `createUser` and `updateUser` keys; `createUser` has a `required` key nested inside with an array of fields, but `updateUser` *doesn't*.  That's because the required fields are only required when creating a new record; when updating, all the fields are optional.  So if you use your `payment` definition for both `create` and `update` it'll mark those fields as required for `update` too.

spec/support/api/schemas/payments.json
```diff
@@ -0,0 +1,35 @@
...
+      "updated_at"
+    ],
+    "properties": {
+      "agency_id": { "type": "string", "format": "uuid" },
+      "amount_cents": { "type": "integer", "minimum": 0 },
+      "amount_currency": { "type": "string" },
+      "care_finished_on": { "type": "string", "format": "date" },
+      "care_started_on": { "type": "string", "format": "date" },
+      "created_at": { "type": "string" , "format": "date-time" },
+      "discrepancy_cents": { "type": "integer", "minimum": 0 },
```
katelovescode, line 26: Does "minimum" imply it has to have a discrepancy?  There's a difference between a discrepancy not being recorded, and it being 0.  I'd like to retain the `null` state as an option, and I wonder if defining this as a minimum means there has to be a value.

spec/support/api/schemas/payment.json
```diff
@@ -0,0 +1,31 @@
...
+    "care_finished_on",
+    "care_started_on",
+    "created_at",
+    "discrepancy_cents",
```
katelovescode, line 9: Discrepancy fields shouldn't be required

spec/support/api/schemas/payments.json
```diff
@@ -0,0 +1,35 @@
...
+  "required": ["payments"],
+  "items": {
+    "type": "object",
+    "required": [
+      "agency_id",
+      "amount_cents",
+      "care_finished_on",
+      "care_started_on",
+      "created_at",
+      "discrepancy_cents",
```
katelovescode, line 12: discrepancy shouldn't be required.

spec/support/api/schemas/payment.json
```diff
@@ -0,0 +1,31 @@
...
+    "updated_at"
+  ],
+  "properties": {
+    "agency_id": { "type": "string", "format": "uuid" },
+    "amount_cents": { "type": "integer", "minimum": 0 },
+    "amount_currency": { "type": "string" },
+    "care_finished_on": { "type": "string", "format": "date" },
+    "care_started_on": { "type": "string", "format": "date" },
+    "created_at": { "type": "string" , "format": "date-time" },
+    "discrepancy_cents": { "type": "integer", "minimum": 0 },
```
katelovescode, line 23: See my other comment about the "minimum" value here.

spec/requests/api/v1/payments_spec.rb
```diff
@@ -0,0 +1,275 @@
...
+      "agency_id": agency_id,
+      "amount_cents": '123400',
+      "care_finished_on": '2020-06-01',
+      "care_started_on": '2020-01-01',
+      "discrepancy_cents": '7890',
+      "paid_on": '2020-07-07',
+      "site_id": site_id
+    }
+  end
+
```
katelovescode, line 21: Could we make another set of params with no discrepancy and make sure we still get a 201 on create?

db/migrate/20200802222331_create_payments.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.string :slug, null: false
+      t.monetize :discrepancy, default: 0
```
katelovescode, line 9: Can we remove this default so we can retain a null state if the discrepancy hasn't been set at all?  Also I wonder if setting this default is what's marking `discrepancy_cents` and `discrepancy_currency` as `not null` in the schema. 🤔 If you remove this default, can you do `rails annotate_models` and see if these are updated?

weedySeaDragon:
No worries about the comments, @katelovescode .  This is what reviews are for. :-)
Will get to them later today-ish.

weedySeaDragon:
@katelovescode see how it looks now.
swagger_helper.rb:  I learned how to use ` allOf:` so now `createPayment` has the `required` constraint (but `updatePayment` doesn't).

In the API `payments_spec.rb` I added a test as you requested for `create` to ensure we get 201 if we don't specify a `discrepancy`.
I also added a test to `create` to ensure that we get a 422 if we don't specify an `amount`.  (I wanted to be really sure to test that it was _required.)
I also added a test to `update` to ensure that we can successfully pass just 1 value to be updated.

And:
- the payments Factory now puts in a value for `discrepancy` about 20% of the time (so most are nil).  We can remove that and/or adjust the %.
- `payments.json` now refers to `payment.json`.  The duplication pained me.

spec/factories/payments.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :payment do
+    paid_on { Faker::Date.between(from: care_finished_on, to: Time.zone.today) }
+    care_started_on { Faker::Date.backward(days: 365) }
+    care_finished_on { Faker::Date.between(from: care_started_on, to: Time.zone.today) }
+    amount { Faker::Number.between(from: 0, to: 999_999) }
+    # Set a discrepancy value about 20% of the time:
```
katelovescode, line 9: 😍

spec/swagger_helper.rb
```diff
@@ -200,7 +219,17 @@
             }
           }
         }
-      }
+      },
+      createPayment: {
+        allOf: [
```
katelovescode, line 224: 😍

spec/swagger_helper.rb
```diff
@@ -33,6 +33,25 @@
...
+          type: :object,
+          properties: {
+            agency_id: { type: :string, example: '3fa57706-f5bb-4d40-9350-85871f698d52' },
+            amount_cents: { type: :integer, minimum: 0, example: 19_999 },
+            amount_currency: { type: :string, example: 'USD' },
+            care_finished_on: { type: :string, example: '2020-03-15' },
+            care_started_on: { type: :string, example: '2020-01-01' },
+            discrepancy_cents: { type: :integer, example: 456 },
+            discrepancy_currency: {
+              anyOf: [
```
katelovescode, line 46: 😍

db/migrate/20200802222331_create_payments.rb
```diff
@@ -0,0 +1,18 @@
...
+      t.monetize :amount, null: false, default: 0
+      t.string :slug, null: false
+      t.monetize :discrepancy, amount: { null: true, default: nil }, currency: { null: true, default: nil }
+      t.uuid :site_id, null: false
+      t.uuid :agency_id, null: false
+
+      t.timestamps
+    end
+    add_index :payments, %i[site_id agency_id]
+    add_index :payments, :site_id
```
arku, line 16: Since we already have a composite index where `site_id` is the first field, we don't need a single index on it.

weedySeaDragon:
Hrm.  I agree that having the default be 0 but then having a validation that the amount be > 0 is inconsistent.  Zero will 'force' the value to be changed.

If we aren't allowing the value to be null/nil, we do have to have a default amount.  If you don't want it to be zero then what would you like it to be? :-)

katelovescode:
> If we aren't allowing the value to be null/nil, we do have to have a default amount. If you don't want it to be zero then what would you like it to be? :-)

This is an interesting statement to me.  I wouldn't expect a default to be required because then it would never hit a "null" condition; the db would always assign a value and the only thing the null constraint would be doing is preventing an EDIT to the field to change it to null.  Can you give me some more info on "we do have to have a default amount"? Is that a monetize thing?

weedySeaDragon:
My perspective is: If you declare the constraint "amount cannot be null" then you are declaring that any Payment with an amount that _is_ null is in an invalid state. And you should never initialize the state of any object/data such that it is invalid.  
Why would you immediately violate the constraints you declared?  If you're gonna immediately violate them, then are you really modeling the constraints correctly?  And if you immediately violate the constraints, it is not just unclear to future readers/developers, it is downright confusing.

I do understand that it can definitely be _implemented_  as you're saying.  (set up the `amount [not null]` constraint in the db and in the Rails model validations, but still allow a new Payment to have a null amount as its initial state)

PR #175, opened 2020-08-08
Feature/159/add sites and child sites

katelovescode:

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally

```
rails db:migrate
```

## 🧳 Steps to test
- hit the /sites API endpoints manually using postman or another tool - you will need to log in first and send your token in the authorization header
- hit the /child API endpoints for update/create using postman or another tool, with nested child_sites attributes (see the Swagger documentation at `localhost:3001/api-docs` for examples - you will need to log in first and send your token in the authorization header

## 🕯 Caveats, concerns

This PR really makes me want to refactor Swagger to 3.0 and work on getting auth into the Swagger endpoints.  :/


app/models/child.rb
```diff
@@ -3,19 +3,23 @@
...
   belongs_to :user
+  has_many :child_sites, dependent: :destroy
+  has_many :sites, through: :child_sites
 
   validates :active, inclusion: { in: [true, false] }
   validates :date_of_birth, presence: true
   validates :full_name, presence: true
   validates :full_name, uniqueness: { scope: %i[date_of_birth user_id] }
 
   validates_each :date_of_birth do |record, attr, value|
```
weedySeaDragon, line 14: Date validation is being done in multiple models with the same code.
I'd suggest creating a custom validator class and then using it in all the places.

app/models/child_site.rb
```diff
@@ -0,0 +1,35 @@
...
+
+# The sites where a child receives care
+class ChildSite < ApplicationRecord
+  # Handles UUIDs breaking ActiveRecord's usual ".first" and ".last" behavior
+  self.implicit_order_column = 'created_at'
+
+  belongs_to :child
+  belongs_to :site
+
+  validates_each :started_care, :ended_care do |record, attr, value|
```
weedySeaDragon, line 11: Another place where a custom date validator can be used.

app/models/site.rb
```diff
@@ -0,0 +1,44 @@
...
+  has_many :children, through: :child_sites
+
+  validates :active, inclusion: { in: [true, false] }
+  validates :name, presence: true, uniqueness: { scope: :business_id }
+  validates :address, presence: true
+  validates :city, presence: true
+  validates :state, presence: true
+  validates :zip, presence: true
+  validates :county, presence: true
+
```
weedySeaDragon, line 19: Should there be a validator for `qris_rating`?  I noticed that the Factory for it just creates values `1..5` -- so is that the valid range?

katelovescode: I'm not sure about the business logic here; QRIS rating might be 1-5 in every state or it might be different; I think we want to prevent a validation because of the likelihood of us running across a state that does their QRIS ratings by letters or some other factor; I'll check with Chelsea though, good catch!

katelovescode:
> 2 suggestions:
> 
> 1. create and use a custom Date Validator
> 2. should site `qris_rating` be limited and validated to certain values?

Resolved 1, added a custom Date Validator.  Getting confirmation on 2

app/models/site.rb
katelovescode: Screen Shot 2020-08-10 at 9 08 35 AM
Yep, looks like they're all gonna be different 😭

Issue #179, opened 2020-08-10
DRY API specs


weedySeaDragon:
## 📋 Summary

The API tests for Rails models are repetitive and use the same patterns for pretty much all of the tests.

## 💻 Implementation Details

We can use `RSpec.shared_examples` and pass parameters (e.g. the class name ['Payment'] and perhaps the lower-class name [`payment`] that can also be pluralized to construct the path [`/api/v1/payments`] etc.

## 🧐 Why tho

1.  We can ensure that all models implement the APIs correctly and in a standardized way.  (Although if there's some model that has a reason to do things differently, we can just write a spec specifically for it.)
2. Changes to how we handle authorizations, changes to the API version path, etc., can then be done in just 1 place.
3.  I cannot stand repetitive, non-DRY code.  It hurts my little software-engineering head.

Spoiler alert:  I've already started this and will submit a PR in a day or so.

PR #181, opened 2020-08-11
#179 RSpec.shared_examples for Rails model API actions

weedySeaDragon:

#179 DRY API Specs  Use RSpec.shared_examples for the common API requests for Rails models.  Pass parameters to the shared_examples as needed to specify names, classes, data to be updated, etc.

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🕯 Caveats, concerns

The shared examples show how much duplication there is in the specs.  Just the names of the classes (and the data to change with the `update` action) vary.  I am passing in the lower case 'item name,' but we could do away with that, since it's based on the class name.

This PR is here so it can be discussed/reviewed. 
There are likely things that can be further refactored/cleaned up/improved; I wanted to get this pushed so we can at least think about it and discuss.

I'm not happy with the names of the shared examples, but those can be changed later.

### DIFF
--- a/spec/support/shared_examples_api_request.rb
+++ b/spec/support/shared_examples_api_request.rb
@@ -13,7 +13,7 @@ VALID_API_PATH = '/api/v1'
 # Ex: Assume you are testing the API calls for creating a Payment. The
 #     item_params are the parameters needed to create a Payment.
 #
-#      it_behaves_like 'it creates an item for a user', Payment, 'payment' do
+#      it_behaves_like 'it creates an item', Payment, 'payment' do
 #        let(:item_params) {
 #          {
 #            "agency_id": agency_id,
@@ -200,7 +200,7 @@ end
 #      and as part of the schema name in the schema definitions.
 #      It is pluralized and used for the path and the tags.
 #
-RSpec.shared_examples 'it creates an item for a user' do |item_class, item_name|
+RSpec.shared_examples 'it creates an item' do |item_class, item_name|
   item_plural = item_name.pluralize
   item_name_symbol = item_name.to_sym
 
@@ -253,7 +253,7 @@ end
 #    update_valid_value [String | Number | nil] - valid value for the updated value for the attribute
 #    update_invalid_value  [String | Number | nil] - invalid value for the attribute so that the server returns a 422 (cannot be updated) error
 #
-RSpec.shared_examples 'it updates an item with a slug, for a user' do |item_class, item_name, update_attribute, update_valid_value, update_invalid_value|
+RSpec.shared_examples 'it updates an item with a slug' do |item_class, item_name, update_attribute, update_valid_value, update_invalid_value|
   item_plural = item_name.pluralize
   item_name_symbol = item_name.to_sym
 
--- a/spec/support/shared_examples_api_request.rb
+++ b/spec/support/shared_examples_api_request.rb
@@ -112,11 +112,16 @@ end
 
 # ------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


spec/support/shared_examples_api_request.rb
```diff
@@ -0,0 +1,343 @@
...
+#
+#  These are the parameters passed to this example:
+#    item_class [Class] - the class for the item; is used to create a new item
+#      with the item_params, and to get the slug for the created item.
+#    item_name [String] - the name of the item (singular).
+#      It is used as a key in the parameters sent to the server
+#      and as part of the schema name in the schema definitions.
+#      It is pluralized and used for the path and the tags.
+#
+RSpec.shared_examples 'it creates an item for a user' do |item_class, item_name|
```
katelovescode, line 203:
```suggestion
RSpec.shared_examples 'it creates an item' do |item_class|
```

spec/support/shared_examples_api_request.rb
```diff
@@ -0,0 +1,343 @@
...
+  item_plural = item_name.pluralize
```
katelovescode, line 204:
```suggestion
  item_name = item_class.class.name.downcase
  item_plural = item_name.pluralize
```

spec/support/shared_examples_api_request.rb
```diff
@@ -0,0 +1,343 @@
...
+#      with the item_params, and to get the slug for the created item.
+#    item_name [String] - the name of the item (singular).
+#      It is used as a key in the parameters sent to the server
+#      and as part of the schema name in the schema definitions.
+#      It is pluralized and used for the path and the tags.
+#    update_attribute [String] - attribute name to be updated
+#    update_valid_value [String | Number | nil] - valid value for the updated value for the attribute
+#    update_invalid_value  [String | Number | nil] - invalid value for the attribute so that the server returns a 422 (cannot be updated) error
+#
+RSpec.shared_examples 'it updates an item with a slug, for a user' do |item_class, item_name, update_attribute, update_valid_value, update_invalid_value|
```
katelovescode, line 256:
```suggestion
RSpec.shared_examples 'it updates an item with a slug' do |item_class, update_attribute, update_valid_value, update_invalid_value|
```

spec/support/shared_examples_api_request.rb
```diff
@@ -0,0 +1,343 @@
...
+  item_plural = item_name.pluralize
```
katelovescode, line 257:
```suggestion
  item_name = item_class.class.name.downcase
  item_plural = item_name.pluralize
```

katelovescode: I'd love to see this instead of the item name passed to every one of these shared examples.  I didn't mark them all up, just the first few examples.

weedySeaDragon: I definitely thought about this too.  Didn't know how much I could really assume about the item name: would it always be based on the class name?  I couldn't think of a situation where it would _not_ be, but didn't really know.
Since you've confirmed that we can assume that -- I'll definitely make the change.  Simplifies things. 👍

weedySeaDragon:
This is definitely worth thinking about.  I'll make the change you suggested to `item_name` and also will remove "... for a user" from the names.

Keep percolating on it.

weedySeaDragon:
I also changed the argument to the "it lists all items for a user" to be the class so that is consistent with the other ones.

Issue #156, opened 2020-07-31
Log In link broken on create an account page


csprayregen:
## 🤬 What's wrong?
When I click the link Log In from the create an account page, I am redirected to a page that only says Report Feedback (with a link that does nothing)

## 👉🏼 Steps to Reproduce

<!-- Steps to reproduce the behavior:
1. Go to https://pie-staging.herokuapp.com/signup
2. Click on "Log In"

## 🧥Expected behavior
I should be redirected to https://pie-staging.herokuapp.com/login

## 📸 Screenshots
![Screen Shot 2020-07-31 at 2 49 36 PM](https://user-images.githubusercontent.com/26717304/89071743-ed3e4100-d33c-11ea-883e-aab3d9bfd905.png)

## 💻 Desktop

Please complete the following information if accessing the app from a computer:

- OS: MacOS Catalina v.10.15.4
- Browser: Chrome
- Version 84.0.4147.105 (Official Build) (64-bit)

## 📱Smartphone
n/a

## 🍣 Additional context
n/a

PR #183, opened 2020-08-13
Fix "Rendered more hooks than during the previous render" error

arku:

Fixes https://github.com/pieforproviders/pieforproviders/issues/156

When you go to http://localhost:3000, the `AuthLayout` component renders the `Login` component by calling the function (i.e. `rightColumnContent()`) and a few hooks from the `Login` component are rendered. If you click on the **SIGN UP** link, the `Signup` function is called and it uses more hooks than the `Login` component. This is the reason for the `Rendered more hooks than during the previous render` error.

When you go to http://localhost:3000/signup, the hooks from the `Signup` component will be rendered the first time. If you click on the **LOG IN** link, there'll be fewer hooks rendered and you'll run into the `Rendered fewer hooks than expected. This may be caused by an accidental early return statement` error.

The PR attempts to fix the issue by rendering the component with the `React.createElement` call (i.e. `<ContentComponent />`) — the `AuthLayout` component no longer has to worry about the order of the hooks in the different render. It is delegated to the individual components (`Login` and `Signup`).

## 🍂 Before You Submit

* [ ] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test

1. Run `foreman start` from the root
2. Click on the **SIGN UP** link from the login page
3. Click on the **LOG IN** link from the signup page. They should both render without any errors.

### DIFF
--- a/client/src/_shared/AuthLayout.js
+++ b/client/src/_shared/AuthLayout.js
@@ -3,7 +3,10 @@ import PropTypes from 'prop-types'
 import piefulltanlogo from '_assets/piefulltanlogo.svg'
 import '_assets/styles/layouts.css'
 
-export function AuthLayout({ backgroundImageClass, rightColumnContent }) {
+export function AuthLayout({
+  backgroundImageClass,
+  contentComponent: ContentComponent
+}) {
   return (
     <div className="grid grid-cols-1 medium:grid-cols-8 large:grid-cols-2 h-screen">
       <div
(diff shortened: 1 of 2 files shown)


client/src/_shared/AuthLayout.js
```diff
@@ -3,7 +3,10 @@ import PropTypes from 'prop-types'
 import piefulltanlogo from '_assets/piefulltanlogo.svg'
 import '_assets/styles/layouts.css'
 
-export function AuthLayout({ backgroundImageClass, rightColumnContent }) {
+export function AuthLayout({
+  backgroundImageClass,
+  rightColumnContent: Component
```
arku, line 8: @katelovescode Let me know if we should change the prop name.

katelovescode:
@arku Can you write an explanation of why this fix works?  Also I'd love it if we could change the name to ContentComponent if possible.

Issue #144, opened 2020-07-10
As a user, I can see that my child roster spreadsheet was uploaded successfully


csprayregen:
## ✨ What new feature do you want to see?
After I finish my spreadsheet upload, I see confirmation and instructions for what to do next. Otherwise, I see an error state. 

## 🧶 Acceptance Criteria
- [ ] After I complete the upload, I see the "spreadsheet uploaded" success message from the mockup
- [ ] After I complete the upload, I can click the "Next" button and be redirected
- [ ] The above criteria work in both English and Spanish

## 🖍️ Implementation Details
Currently we just display the data after the upload, now we want to build out the user's onboarding flow, and where they are directed after upload

Upload details are covered in #36 and #116. Error messages for this screen are covered in #402.
An upload bug is covered in #84 
Upload instructions are covered in #71 and #398 

## 🥧 Additional context
Mockup (tablet and desktop only - mobile unavailable): https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=281%3A1

csprayregen:
@rebeccakarasiktw I can't see these messages anywhere in Figma?  Did we decide to nix this part? Am I losing my mind?

PR #189, opened 2020-08-18
Support XLSX and XLS files upload

arku:

Fixes https://github.com/pieforproviders/pieforproviders/issues/84

- Supports uploading of XLS and XLSX files
- Updates the route to `/cases/import` as `/csv-import` no longer makes sense. And replaces references to `CSVImport` with `CasesImport`

Screenshots

image

image

## 🍂 Before You Submit

* [ ] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally

Run `cd client && yarn` to install the dependencies

## 🧳 Steps to test

1. Log in and go to `/cases/import`
2. Upload an XLS cases file with data. You should find the data displayed in a table
3. Repeat Step 3 for XLSX and CSV files

## 🕯 Caveats, concerns

The code is tightly coupled to the sample sheet structure provided in https://github.com/pieforproviders/pieforproviders/issues/144. The code skips the first 3 rows as they are filled with instructions, headers, and column descriptions. If the template changes, we'll have to update the code. If a user removes these rows or changes the column order, the code might not work as expected. I would recommend moving the instructions and the column description rows to the application if possible.

We might want to use [React Virtualized](https://github.com/bvaughn/react-virtualized) or [React Window](https://github.com/bvaughn/react-window) if we anticipate too many rows in the uploaded file.


katelovescode:
@csprayregen couple of thoughts on this that I'd like to run by you tonight during the working session, but see Arun's notes above:

This manually removes the first 3 rows because that's what our sample sheet has for instructions, etc.  That means if anyone tries to upload something NOT in our template, it won't parse correctly.  Also if the columns are out of order, same deal.  We have talked a couple times about whether or not we're going to require users to use our template, and I don't remember where we landed.  If the answer is "you must use our template" we might be able to put in some checks, i.e. "if the first cell doesn't say 'Pie for Providers' sample, give the user an 'invalid format, please use the sample spreadsheet' error".

PR #187, opened 2020-08-15
Set up i18n in the frontend

arku:
# 💅🏼 What issue does this fix?

Fixes #23 

- Sets up [react-i18next](https://react.i18next.com/)
- Translates the content on the Login page.

**Screenshot**

English
image

Spanish
image

## 🍂 Before You Submit

* [ ] Did you write tests?
* [ ] Did you run `rails rswag` from the root?
* [ ] Did you run `rubocop` from the root?
* [x] Did you run `yarn lint` in `/client`?
* [x] Did you run `yarn test` in `/client`?
* [ ] Are your primary keys UUIDs on any new tables?

## 🛷 Deployment Considerations

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally

Run `cd client && yarn` to install the i18next packages

## 🧳 Steps to test

1. Get the app up and running and visit the login page
2.  Change the language in the switcher on the top-right. You should find the page content translated.

Issue #23, opened 2020-02-22
Internationalization


katelovescode:
## ✨ What new feature do you want to see?
The app copy can be read in both English and Spanish

## 🏖 Implementation Details
Most of the copy we're concerned with will be on the front-end side so this ticket should:
- find an internationalization library for react
- talk to @rebeccakarasiktw and @csprayregen for translations or next steps to get translations

Stretch goal:
- either translate all the existing copy in the app, OR
- write a ticket for that translation with a full audit of currently-implemented copy as part of the ticket description

katelovescode:
* [ ]  go to any page, and click the language switcher at the top to change to Spanish. Make sure the language changes for the form labels, etc.
>   This works for sign up and log in. For "thanks for signing up" the language button doesn't do anything. There is no language button on the Getting Started page. @katelovescode do you prefer I write separate tickets for these?

This will get a new ticket, Chelsea's got it on her todo list

> * [ ]  submit a form with failing fields, and make sure the errors are in Spanish
>   This works partially. See two English errors in the login page:

@arku are these two backend errors? (I think so) so handled by #185 

> No errors are produced when I fail to sign the terms of use unless it is the only thing wrong - I couldn't get it to appear here.

This requires a separate ticket, the issue is that the validation will only work the first time, if you un-check after it's passed validation it will not show a validation error.

> Other English error messages here on signup:

@arku pretty sure these are also backend, so handled by #185 

> * [ ]  visit another page (by logging in or navigating via a link) and make sure the language stays the same - this did work. However, the words Setup and Steps are not translated.
>   ![Screen Shot 2020-08-22 at 8 32 16 PM](https://user-images.githubusercontent.com/26717304/90968940-13787c00-e4b8-11ea-9904-70ed37741bd3.png)

@arku can you fix this pre-QA? (setup and steps)

> * [ ]  switch back to English and make sure the language changes and the same behaviors can be reproduced in English - this did not work after log in, since there was either no button or a broken button to switch languages. It did work on the sign up and log in page. However, errors stay in Spanish when I switch back to English.
>   ![Screen Shot 2020-08-22 at 8 55 40 PM](https://user-images.githubusercontent.com/26717304/90969069-c85f6880-e4b9-11ea-9922-0f9fd321a60c.png)

@arku this should probably be addressed before re-QA

csprayregen:
@arku @katelovescode I'm still seeing the same errors noted above - moving this back to In Progress:

[...]

PR #193, opened 2020-08-20
Change dropdown to link, additional translation, some cleanup of form elements

katelovescode:
Related to #23 and #187 - changes the dropdown language selector to a link per the mockups, adds additional translation in other pages, and cleans up unused custom form elements

## 🍂 Before You Submit

* [ ] Did you write tests?

## 🧵 Steps to set up locally

Pull down changes, if you have #187 you will only need to let it re-render

## 🧳 Steps to test
Make sure the language switcher is a button and that clicking on it changes the language.  When you're in "English", the button should say "Español", and vice versa.  Some translations are available for the signup page; when you go to signup, and change to Spanish, you'll see SOME fields change.  Also changing between pages should retain your language preference.

## 🕯 Caveats, concerns
Rendering issues in a render test, cc @arku - check Slack for details


client/src/Signup/Signup.js
```diff
@@ -230,7 +227,7 @@ export function Signup() {
               setUser({ ...user, language: event.target.value })
```
arku, line 227: **suggestion (non-blocking)**: What do you think of setting the initial value of `language` [here](https://github.com/pieforproviders/pieforproviders/pull/193/files#diff-1f11d77c4fc8b72a41680a8f3e750206R23) to `i18n.language`?

client/src/i18n/locales/es/index.json
```diff
@@ -3,13 +3,60 @@
...
   "email": "Correo",
   "password": "Contraseña",
   "emailAddressRequired": "Correo electrónico es requerido",
   "passwordRequired": "Contraseña es requerido",
   "forgotPassword": "¿Olvidaste tu contraseña?",
   "resetPasswordText": "No te preocupes. Haz clic en el botón abajo y revisa tu correo para resetearla.",
   "resetPassword": "Resetea tu contraseña",
-  "pieforProvidersLogoAltText": "Logo de Pie for Providers"
+  "pieforProvidersLogoAltText": "Logo de Pie for Providers",
+  "agreeToTerms": "He leído y estoy de acuerdo con",
```
arku, line 17: Missing space at the end.

image

client/src/_shared/ActionLink.js
```diff
@@ -0,0 +1,27 @@
...
+      >
+        {text}
+      </Button>
+    </>
+  )
+}
+
+ActionLink.propTypes = {
+  onClick: PropTypes.func,
+  text: PropTypes.string,
```
arku, line 25: Can we mark these props as required?

client/src/_shared/ActionLink.js
```diff
@@ -0,0 +1,27 @@
...
+export function ActionLink({ onClick, text, classes }) {
+  const handleClick = e => {
+    e.preventDefault()
+    onClick()
+  }
+  return (
+    <>
+      <Button
+        onClick={handleClick}
+        className={`${classes} focus:shadow-none`}
```
arku, line 14: We might want to check if `classes` is truthy as it is an [optional](https://github.com/pieforproviders/pieforproviders/pull/193/files#diff-a467d5fddf5ec774cd44e196a71c6887R26) prop or set a default value.

PR #220, opened 2020-08-24
Improve signup error handling

arku:

- Only sets validation errors if the response status is 422
- Introduces a new state for non-validation errors to capture 500 responses from the server and displays a generic error message.
- Resets the error state before submitting the form

## 🍂 Before You Submit

* [ ] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test

Sign up with an existing phone number or email and you should find the error message show up.

### DIFF
--- a/client/src/Signup/Signup.js
+++ b/client/src/Signup/Signup.js
@@ -1,6 +1,6 @@
 import React, { useEffect, useState } from 'react'
 import ReactGA from 'react-ga'
-import { Form, Input, Select, Radio, Checkbox } from 'antd'
+import { Alert, Form, Input, Select, Radio, Checkbox } from 'antd'
 import { PaddedButton } from '_shared/PaddedButton'
 import { Link } from 'react-router-dom'
 import MaskedInput from 'antd-mask-input'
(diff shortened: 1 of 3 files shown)


client/src/i18n/locales/es/index.json
```diff
@@ -59,5 +59,7 @@
   "gettingStartedWelcome": "Bienvenidas a Pie for Providers, ",
   "gettingStartedTitle": "Comencemos",
   "gettingStartedInstructions": "Sigue estas instrucciones para configurar tu panel de caso. Esto debería tardar unos 15 minutos y solo tendrás que hacerlo una vez. ¡Prepárate para aumentar su porción del pastel!",
-  "gettingStartedButton": "Comienza"
+  "gettingStartedButton": "Comienza",
+
+  "genericErrorMessage": "Something went wrong, try again later."
```
arku, line 64: @csprayregen Need your help translating this piece of text.

client/src/Signup/Signup.js
```diff
@@ -47,10 +51,11 @@ export function Signup() {
     })
     if (response.status === 201) {
       setSuccess(true)
-    } else {
+    } else if (response.status === 422) {
       const { errors } = await response.json()
-      setErrors(errors[0].detail)
-      // TODO: Sentry
```
arku, line 53: This comment is removed because we'll be capturing the exceptions from the backend.

client/src/i18n/locales/es/index.json
katelovescode, line 64:
```suggestion
  "genericErrorMessage": "Algo salió mal, inténtalo de nuevo más tarde."
```

katelovescode:
This is approved and I did translation live with Chelsea.

Issue #160, opened 2020-08-02
DB Implementation: Case Cycles & Child Case Cycles


katelovescode:
## 📋 Summary
Write the migration, model and controller actions necessary for basic CRUD on Case Cycles from the data model.

## 💻 Implementation Details
- This is an authenticated route
- Users will be managing case cycles on their account and will need authorization (i.e. they can only add/edit/update/delete case cycles associated to their account)
- All CRUD actions are required
- see docs/dbdiagram.dbml for field definitions and associations
- this work should include creating child_case_cycles as an association table, as children are already implemented as a model
- technically this work is blocked by #161 but the association can probably be made as a part of that ticket if this work comes first; if you pick up this ticket and 161 has no merged PR, please update that ticket to ensure the association is made

## 🧐 Why tho
Start getting the DB built to make front end tickets easier

arku:
Done in #228, #293, #311, and #324.

PR #228, opened 2020-08-24
Set up CaseCycle model

arku:

#160

- Creates a `CaseCycle` model. Uses Postgres enums (along with ActiveRecord enums) for `status` and `copay_frequency` columns to avoid storing meaningless integers in the table and to maintain data integrity. Check https://sipsandbits.com/2018/04/30/using-database-native-enums-with-rails for more information on the topic.

- Fixes `DateParamValidator` to add errors to the attributes. It was previously adding to `:date_param` regardless of the validation errors on different date columns.
- Adds specs for `CaseCycle` model

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally

Run `rails db:migrate`

## 🕯 Caveats, concerns

I had to change the schema format from `:ruby` to `:sql` as it doesn't support the dumping of models with native enums and fails with the following error message:

```
# Could not dump table "case_cycles" because of following StandardError
#   Unknown type 'case_status' for column 'status'
```

Changes:

- The database schema will be tracked in `db/structure.sql`
- We'll have to use `rails db:structure:load` and `rails db:structure:dump` instead of the `db:schema` variants.


app/models/case_cycle.rb
```diff
@@ -0,0 +1,47 @@
...
+  monetize :copay_cents
+
+  enum copay_frequency: { daily: 'daily', weekly: 'weekly', monthly: 'monthly' }
+  enum status: { submitted: 'submitted', pending: 'pending', approved: 'approved', denied: 'denied' }
+
+  before_save :set_slug
+
+  validates :slug, uniqueness: true
+  validates :copay, numericality: { greater_than: 0 }
+  validates :submitted_on, date_param: true, allow_blank: true
```
weedySeaDragon, line 14: add `validates :copay_frequency, inclusion: { in: copay_frequency.keys }`
and `validates :status, inclusion: { in: status.keys }`

spec/models/case_cycle_spec.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe CaseCycle, type: :model do
+  let(:invalid_date_msg) { 'Invalid date' }
+
+  it { should validate_numericality_of(:copay).is_greater_than(0) }
+
```
weedySeaDragon, line 9: Need specs/examples/tests for `status`:

```
it {
    should define_enum_for(:status).with_values(
      submitted: 'submitted',
      pending: 'pending',
      approved: 'approved',
      denied: 'denied'
    ).backed_by_column_of_type(:enum) }

  it { should allow_values(:submitted, :pending, :approved, :denied).for(:status) }

```

spec/models/case_cycle_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+  it 'validates uniqueness of slug' do
```
weedySeaDragon, line 10: Need specs/examples/tests for `copay_frequency` as well

spec/models/case_cycle_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    create(:case_cycle)
+    should validate_uniqueness_of(:slug)
+  end
+
+  it 'validates submitted_on date if present' do
+    cycle = build(:case_cycle, submitted_on: nil)
+    expect(cycle).to be_valid
+
```
weedySeaDragon, line 18: I would add lines to explicitly show that setting `cycle.submitted` to a valid `Date` is valid (`be_valid`).  That will provide an example/specification that shows that a `Date` is the correct thing to use.   (And the same for `effective_on`, `expires_on`, and  `notified_on`

spec/factories/case_cycles.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :case_cycle do
```
weedySeaDragon, line 4: @katelovescode Should the factory randomly set `status`, `copay_frequency`, and/or the dates (`notified_on` etc.) ?
(e.g. by using `sample`)

And or should we define `traits` based on the status?  (Or should we wait until we really need them in other specs?)

app/models/case_cycle.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+# The cycle for subsidy cases
+class CaseCycle < UuidApplicationRecord
+  monetize :copay_cents
+
+  enum copay_frequency: { daily: 'daily', weekly: 'weekly', monthly: 'monthly' }
+  enum status: { submitted: 'submitted', pending: 'pending', approved: 'approved', denied: 'denied' }
+
+  before_save :set_slug
```
weedySeaDragon, line 10: FWIW, I like this style better than using a lambda/block here.  Is cleaner to me.

app/models/case_cycle.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+# The cycle for subsidy cases
+class CaseCycle < UuidApplicationRecord
+  monetize :copay_cents
+
+  enum copay_frequency: { daily: 'daily', weekly: 'weekly', monthly: 'monthly' }
```
katelovescode, line 7: As an alternative, what if we used a regular string field, and validated that the input is included in a constant array of values?

i.e.
```
class CaseCycle

FREQUENCIES = [ "daily", "weekly", "monthly" ]

validates :copay_frequency, inclusion: { in: FREQUENCIES }
```
This gives us the added benefit of using the defined constant outside of the model, so we can use the same list for specs, etc.

Thoughts? @weedySeaDragon @arku ?

weedySeaDragon: That's exactly what an `enum` is for:  so you have a structured, reusable way of saying "here is an enumerated list of things to use; you can only choose one of them".
WRT validations, the _specification_ should not re-use the enums, because then you're not really _specifying_ anything.  (it's kind of circular reasoning instead of explicitly saying "here are the items that can be in that enumerated list.)
But when you're testing the behavior (or internal stuff), it's definitely ok to use the enum itself.

If you need to reuse an `enum` across classes like we do with _license types,_  then we can do this (which I just happened to write a bit ago as I'm working thru the Subsidy Rules):
```
# frozen_string_literal: true

# License types
# These are the list of valid license types.
# These correspond to an ENUM type defined in the Postgres DB
class Lookups::LicenseType
  enum valid_types: { licensed_center: 'licensed_center',
                      licensed_family_home: 'licensed_family_home',
                      licensed_group_home: 'licensed_group_home',
                      license_exempt_home: 'license_exempt_home',
                      license_exempt_center: 'license_exempt_center' }
end
```

Here's an example of how it can be used in the Business class:
```
class Business
   ...
   enum license_types: Lookups::LicenseType.valid_types
  ...

```

spec/factories/case_cycles.rb
arku: @katelovescode What's the default status of a `CaseCycle`?

app/models/case_cycle.rb
arku: When it is present, the spec `it { should validate_inclusion_of(:status).in_array(CaseCycle::STATUSES) }` fails because when you try to initialize an object with an invalid status (`'test'`, for instance), ActiveRecord enum throws an `ArgumentError ('test' is not a valid status)`. Check https://github.com/rails/rails/issues/13971 for more information.

spec/models/case_cycle_spec.rb
arku: @weedySeaDragon Isn't the second test superfluous? 🤔

arku:
@weedySeaDragon I have addressed most of your comments. Do take another look when you get a chance.

Thanks in advance!

spec/factories/case_cycles.rb
katelovescode: Oop yes, good catch, can you fix that?

spec/models/case_cycle_spec.rb
weedySeaDragon: Agreed.  Some folks like to have both:  they _specify_ slightly different things.  Even though I'm quite on the "these are _specifications!!_" side of things, there's not much real additional value to the 2nd; the values in an :enum also define the allowed ("enumerated") values. 
(I just copied and pasted from existing tests.  I should have actually reviewed what I pasted more carefully.)

app/models/case_cycle.rb
weedySeaDragon: I think we covered this in Slack, but just to follow up here:  What you're writing as 'tests' are actually also _specifications._   Yes, you may be writing unit tests, but they also serve as a description/specification of what should happen.  (The "Spec" in "RSpec" means _specification_.  I can tell you the whole history sometime.  Spoiler: "xSpec" originates from a Smalltalk person: Kent Beck).

Someone is going to later read through the specs/tests to help them understand what is going on and how it is all supposed to be working.  (Code is written for people; bits are for computers.)

I do also agree that it's always a balance with design:  you need to make sure that future (and current!) developers will be able to maintain the code in a reasonable way.  With FOSS, the knowledge varies so _widely._

spec/factories/case_cycles.rb
katelovescode: I updated and pushed the diagram

app/models/case_cycle.rb
weedySeaDragon: <sigh>  Wish they (rails) would re-open that issue and make sure it's addressed.

Issue #237, opened 2020-08-25
db:seed causes email to be created (and letter opener displays it)


weedySeaDragon:
## 🤬 What's wrong?

Do we really need emails to be 'sent' when a new User is created during db:seed?  
We should suppress it.  (Also, letter opener opens it up in the browser. It is just 1 user right now, so is just moderately annoying but in the future...)

## 👉🏼 Steps to Reproduce
(with empty/truncated db tables) run `bundle exec rails db:seed`

## 🧥Expected behavior

We should discuss.  Is creating the emails OK or not?

## 💻 Desktop

Please complete the following information if accessing the app from a computer:

## 📱Smartphone

Please complete the following information if accessing the app from a mobile phone:

## 🍣 Additional context

Tiny problem.  Just need to decide what is the 'correct' behavior.

weedySeaDragon:
Close by #240

PR #240, opened 2020-08-26
Turn off email deliveries when seeding the database

arku:

https://github.com/pieforproviders/pieforproviders/issues/237

And confirms the user account created from the seeds.

## 🍂 Before You Submit

* [ ] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test

`rails db:drop && rails db:setup`

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+ActionMailer::Base.perform_deliveries = false
+
 puts 'seeding'
 
 user = User.where(email: ENV.fetch('TESTUSER_EMAIL', 'test@test.com')).first_or_create(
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+ActionMailer::Base.perform_deliveries = false
```
weedySeaDragon, line 3: Do we want to only do this if it is the development environment? 
@katelovescode -- do you use db:seed with staging, for example?

katelovescode: I don't seed staging, no.  I think we're okay to do this no matter what.  If we end up seeding and/or re-seeding staging, we probably don't want emails to go out anyway.

db/seeds.rb
```diff
@@ -19,6 +21,8 @@
   timezone: 'Central Time (US & Canada)'
 )
 
+user.confirm
```
katelovescode, line 24: 👍

Issue #197, opened 2020-08-21
DB implementation: US address components (states, counties, cities)


weedySeaDragon:
## 📋 Summary

There are address components that are essentially 'lookup' lists:  
- states
- counties
- cities

(We need a better title for this issue. Someone please improve it!)

We also need to decide how to handle when the user wants to enter something  _not_ already in a list.  

## 💻 Implementation Details

- where do we get the data?  what is the _definitive source?_
   - candidate:  `my_zipcode` gem https://github.com/midwire/my_zipcode_gem  or  https://github.com/midwire/free_zipcode_data  (would one or both of these two work?)
   - sources for raw data that we'd need to process:
      - candidate: US Census.gov https://www.census.gov/library/reference/code-lists/ansi.html
         This page has links for the list of US States and territories with the ANSI FIPS codes and official Postal service abbreviations, and the list of US cities and counties from the Census
      - candidate: USGS list of geographic names https://www.usgs.gov/core-science-systems/ngp/board-on-geographic-names/download-gnis-data

- how do we import the data into the system?
   - rake task?  some other program?
   - cron action that runs a rake task (or whatever we decide)  [use the `whenever` gem?]
- how often do we check the definitive source for updates?
   - states (and US territories) do not change very often, so don't need to check these frequently
   - cities and counties?

There is a gem that 

## 🧐 Why tho

Users should be able to select the component (e.g. state, county, city) from a pre-populated list.

weedySeaDragon:
Chatting today in Slack, we decided:
1. for MVP, we'll read these in from a .csv file (e.g. via rake).   Getting that file updated regularly from an official source can be done later.  (post MVP)
2. we'll store these as ActiveRecords.  There are thousands of counties, more cities, etc.  The db backend will provide good performance.

PR #243, opened 2020-08-27
Add address items used for lookup (states, cities, etc.)

weedySeaDragon:

#197 

This creates 4 "lookup" AR models:  Lookup::State,Lookup::County, Lookup::City, Lookup::Zipcode, and provides the rake tasks and .csv data to populate them.

This doesn't create any Contollers or rswag stuff, and it doesn't connect any of the existing models to these.  I wanted to get this nailed down first, and then we can go from there.

I started out with the code from the  [my_zipcode_gem gem](https://github.com/midwire/my_zipcode_gem) (yes, that's really the name) because it was free and easy.  I took the rake file and then optimized it (cause it was horrendously slow in inefficient).  There's still ugliness in there, but it works.  And we won't be importing stuff very often.

## 🍂 Before You Submit

* [x] Did you write tests?

* [x] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally
1. run `db:migrate` as usual
2. To import data into the files, run the rake task `pie4providers:address_lookups:import`
   - this will read in .csv files from `db/lookup_data` and populate the 4 new lookup tables: `states`, `cities`, `counties`, and `zipcodes`
   - the csv files are from the `free_zipcode_data` gem repo, which got it from https://www.geonames.org/

## 🕯 Caveats, concerns

I did create FactoryBot factories for these, even though we could probably use Faker::Address in most places.  I thought having the factories would be better because they'd be aligned with exactly what we have in our models.


Gemfile
```diff
@@ -5,6 +5,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 ruby '2.6.6'
 
+gem 'activerecord-import'
```
arku, line 8: @weedySeaDragon Just curious, is there a reason why we are not using [`upsert_all`](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-upsert_all) and instead relying on this gem?

weedySeaDragon: Ah.  excellent question.  Just habit. 
I'll play with `upsert_all` today and see if it's cleaner.

weedySeaDragon: So I ran some simple tests.  The performance is pretty much the same between using the `activerecord-import` gem and using `insert_all!`.  
When using `insert_all!` I do have to manually create the array of Hashes for each row that needs to be inserted. (take the column headers and make them the keys for all of the values in a csv row).  `activerecord-import` takes care of that.  
Seems at this point, `activerecord-import` is fine to use with relatively small sizes (thousands),and it does provide a fair amount of validity checks w.r.t. the column names (checking that they work with the attribute names), etc.  We're not using any of that functionality, of course, so it's of no benefit.
(And in the repo for the `activerecord-import` gem, people are definitely asking what the future of it is given `insert_all` and `upsert..` now in Rails 6.)

I'll remove the gem and just use `insert_all!`.  That's 1 less gem that is needed.

Good catch!

lib/tasks/address_lookup_items.rake
```diff
@@ -0,0 +1,355 @@
...
+
+      default_states_source = File.join(DEFAULT_INPUT_DIR, INPUT_FN_STATES)
+      states = CSV.read(default_states_source, headers: true).to_a.drop(1)
+
+      create_objects(klass, STATES_CSV_COLS, states)
+      puts_done_msg
+    end
+
+    # import_all time is about the same
+    desc 'Import counties (assumes states have been imported'
```
katelovescode, line 43:
```suggestion
    desc 'Import counties (assumes states have been imported)'
```

lib/tasks/address_lookup_items.rake
```diff
@@ -0,0 +1,355 @@
...
+    MIN_COUNTIES_EXPECTED = 3000
+
+    STATES_CSV_COLS = %i[abbr name].freeze
+    COUNTIES_CSV_COL_VALUES = %i[name state_id county_seat].freeze
+    COUNTIES_CSV_COL_HEADERS = %w[name state county_seat].freeze
+    ZIPCODES_CSV_COL_VALUES = %i[code city_id state_id county_id area_code lat lon].freeze
+    ZIPCODES_CSV_COL_HEADERS = %w[code city state county area_code lat lon].freeze
+    # -----------------------------------------------------------------
+
+    # import_all is about twice as fast
```
katelovescode, line 28: Can we remove these comments now that benchmarking has been done and is recorded in the PR comments?

weedySeaDragon: oops -- yup. will do.

db/structure.sql
```diff
@@ -155,11 +182,62 @@ CREATE TABLE public.children (
 
 
 --
--- Name: data_migrations; Type: TABLE; Schema: public; Owner: -
+-- Name: lookup_cities; Type: TABLE; Schema: public; Owner: -
 --
 
-CREATE TABLE public.data_migrations (
-    version character varying NOT NULL
```
katelovescode, line 162: I'm a little unclear why the data migrations table would come out of the schema with these changes, can you investigate?

weedySeaDragon: Aha.  I didn't realize we needed to be using `db:migrate:with_data`  I was missing the **with_data** part.

In another project I've worked on, we've written a task that calls all of the steps to:
- drop the db (if they exist)
- create the db
- db:migrate
- seed

So there's just 1 task to always run in development.  
We could write this task and then be sure it uses `db:migrate:with_data`
(as a separate PR).

katelovescode: Feel free to add that task if you like!

Issue #246, opened 2020-08-28
DB: Add license types (enum)


weedySeaDragon:
## 📋 Summary

'license type' should be an enum so that it can be used in `business` and `subsidy rule`

## 💻 Implementation Details

1. Add as an ENUM type in the db
2. Add as a simple model (not an ActiveRecord)

## 🧐 Why tho

Needs to be used by multiple models

PR #252, opened 2020-08-28
Add LicenseTypes to be used as a common enum

weedySeaDragon:

#246 

Adds a class that can be used where-ever a `license_type` enum is needed.  (Business, SubsidyRule)
This  has the LicenseTypes class, a method in the ApplicationController to handle translation of this (and any other) enum in a model, and placeholder entries in the `config/locale` files

This PR doesn't include any actual use. (Future PRs should do that) But it would look like this:
```
class Business
     enum category: LicenseTypes.valid_types
   ...
```
Once we settle on this, we can create an issue to update `Business` to use this.  (Spoiler:  I've pretty much got it done; needed to verify this worked.)

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally

Before running tests, be sure to run  `bundle exec rails db:structure:dump` and then  `bundle exec rails db:test:load_structure`
(as @arku said in #228,  we have to switch to using the db structure file so that we can use enums).

## 🕯 Caveats, concerns

1. Do we want the class method that returns the enums Hash to be named `valid_types` ?
    It's kind of repetitive.   Is there a more generalized name we can use and standardize it for all shared enums?  (`.list`? `.enums`?)

2. I did not put this under the `lookups` directory.  I went back and forth on it.  I left it with the other models because although this is a list used  as a  _look up,_  the values are part of the PieForProviders domain.  
I could go either way.


config/locales/es.yml
```diff
@@ -1,4 +1,16 @@
 es:
+  license_types: &license_types
```
katelovescode, line 2: @csprayregen can we get translation on these?

katelovescode:
I also agree that because it's more business domain'y, that it should stay out of the lookups. Good choice.

weedySeaDragon:
> I'd love to change the model to `Licenses` and the method to `.types`, does that make sense?

Makes perfect sense. :-)  Will do.

config/locales/es.yml
```diff
@@ -1,4 +1,16 @@
...
+    licensed_center: licensed center
+    licensed_family_home: licensed family home
+    licensed_group_home: licensed group home
+    license_exempt_home: license exempt home
+    license_exempt_center: license exempt center
```
csprayregen, line 7:
```suggestion
    license_exempt_center: centro exento de licencia
```

config/locales/es.yml
```diff
@@ -1,4 +1,16 @@
 es:
+  license_types: &license_types
+    licensed_center: licensed center
+    licensed_family_home: licensed family home
+    licensed_group_home: licensed group home
```
csprayregen, line 5:
```suggestion
    licensed_group_home: guardería de grupo en el hogar con licencia
```

katelovescode (dismissed):

katelovescode (approved):

PR #256, opened 2020-08-30
Only run E2E tests on pushes and PRs to the main branch

arku:

From the discussion on Slack, we have decided to only run the Cypress E2E tests on pushes and pull requests to the main branch.

## 🍂 Before You Submit

* [ ] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies


katelovescode:
I have since changed my mind.  Let's do this and when we can reduce the speed later on we can revisit

Issue #154, opened 2020-07-22
Implement layout for onboarding pages accessible after login


cyranix:
## 📋 Summary

Add a reusable page layout that applies to pages accessible by a logged in user. 

## 💻 Implementation Details

See Figma screens under "Account Ledger / Add a Payment" for (what I believe to be) the most complete design for the layout, which includes
* top navbar with 
  * white background
  * Pie logo and "Pie for Providers" title on left side
  * user profile picture on right side, which links to the profile page
  * navigation links for "Ledger", "Attendance", and "Search Cases"
    * which collapse behind a hamburger menu when the user is on a mobile device
* outermost content container with 
  * `#D4D4D4` background color
  * interior padding
* design components included:
  * Ant breadcrumb for "Setup"
  * Tailwind base shadow for header & main content

I'd like to ask for some extra guidance from a P4P designer to review the above. 

Additionally, I'm seeing a few different styles for the "page title" located between the top navbar and the main content area: breadcrumbs, larger blue text, smaller dark-gray text, and absent. Are these intended to be different based on viewport size, or more consistent? (This might not end up being part of the layout but could be useful context.)

## 🧐 Why tho

Spun out of #65 because I'm trying not to block development. I imagine that the onboarding workflow can continue to be developed (at least from a functional perspective) independently of adding the outermost page chrome.

Following the merge of #148, please apply this layout to that page. 🙇

csprayregen:
top navbar with
- [ ] white background
- [ ] Pie logo and "Pie for Providers" title on left side
- [ ] user profile picture on right side, which links to the profile page
- [ ] navigation links for "Ledger", "Attendance", and "Search Cases"
- [ ] which collapse behind a hamburger menu when the user is on a mobile device
- [ ] outermost content container with
- [ ] #D4D4D4 background color
- [ ] interior padding
design components included:
- [ ] Ant breadcrumb for "Setup"
- [ ] Tailwind base shadow for header & main content

PR #259, opened 2020-08-31
Issue 154

jontrainor:

these two issues:
 Pie logo and "Pie for Providers" title on left side
The words "Pie for Providers" should disappear in tablet and mobile, but they do not

Tailwind base shadow for header & main content
the shadows should follow what is here: https://tailwindcss.com/docs/box-shadow#app
but it looks like 1- the header has no shadow
2- the card has the shadow-sm applied to it. it should have the shadow (regular) style applied to it.

from  the qa comments in #154 

image

image

## 🍂 Before You Submit

* [ ] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test

👀 the `/getting-started` page to see the logged in layout

## 🕯 Caveats, concerns

The shadow was using `default` and i bumped it up to `md` from these [default configs](https://github.com/tailwindlabs/tailwindcss/blob/a2956f8a75dec0047764abddc7ca490a2ca4596c/stubs/defaultConfig.stub.js#L203) but we can make that bigger still if you want.

### DIFF
--- a/client/src/_shared/LoggedInLayout.js
+++ b/client/src/_shared/LoggedInLayout.js
@@ -2,13 +2,16 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { useHistory } from 'react-router-dom'
 import pieSliceLogo from '_assets/pieSliceLogo.svg'
-import { Breadcrumb, Button } from 'antd'
+import { Breadcrumb, Button, Grid } from 'antd'
 import '_assets/styles/layouts.css'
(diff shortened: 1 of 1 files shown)


jontrainor:
@katelovescode check this out:

image
image

i think that's better but would love a second set of 👀 !

katelovescode:
Looks great!  Nice work.

PR #275, opened 2020-09-02
Add not null constraint to users' greeting name

arku:
# 💅🏼 What issue does this fix?

The greeting name has a presence validation in the frontend but the table lacks it. This PR adds a not-null constraint to `users.greeting_name` and adds a model validation.

## 🍂 Before You Submit

* [x] Did you write tests?
* [x] Did you run `bundle exec rspec` from the root?
* [x] Did you run `bundle exec rails rswag` from the root?
* [x] Did you run `bundle exec rubocop` from the root?
* [ ] Did you run `yarn lint` in `/client`?
* [ ] Did you run `yarn test` in `/client`?
* [ ] Are your primary keys UUIDs on any new tables?

## 🛷 Deployment Considerations

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally

Run `rails db:migrate`

PR #276, opened 2020-09-02
Add user_id to case cycles table

arku:

- Adds a `user_id` column to the `users` table.
- Adds associations to the `User` and `CaseCycle` models.
- Update `CaseCycle` factory

Blocked by https://github.com/pieforproviders/pieforproviders/pull/275

## 🍂 Before You Submit

* [ ] Did you write tests?

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally

Run `rails d:migrate`


weedySeaDragon:
Since this isn't linked to an issue that describes _why_ I think you should put that into the description of this PR.
I think that you should copy the text from what @katelovescode  said in Slack [assuming that's ok with her].

spec/models/case_cycle_spec.rb
```diff
@@ -94,8 +94,14 @@
 #  submitted_on    :date             not null
 #  created_at      :datetime         not null
 #  updated_at      :datetime         not null
+#  user_id         :uuid             not null
```
weedySeaDragon, line 97: Need to add an example (test) for the `belongs_to` relationship, e.g.
```
it { should belong_to(:user) } # or it { is_expected_to belong_to(:user) }
```

weedySeaDragon (changes requested):
just the 1 thing: need an addition to the _spec

spec/models/case_cycle_spec.rb
katelovescode: Concur!

weedySeaDragon (approved):

katelovescode (approved):
https://pie-for-providers.slack.com/archives/C017TFWAJLE/p1598919788004600 discussion for posterity

Relevant part: So yes, in the end, I think we should associate it to users, because the user might want to enter a case cycle and then multiple kids to it.

PR #286, opened 2020-09-05
Feature/84/fix xls csv

katelovescode:
#84 some bugs that seem browser dependent

Sometimes the `type` of an imported XLS file is an empty string, so we're checking both type and file extension

Also, sometimes the index is `undefined` when mapping/reducing the data rows(i.e. if there are 25 columns, we will have 0-24 defined and then an undefined 25th column), so I put a try-clause `?` before `.key` so it won't choke on `.key undefined method for nilClass`

## 🍂 Before You Submit

* [x] Did you write tests?

## 🧳 Steps to test
- visit `/cases/import`
- upload the test csv, xls, and xlsx files
- you should see the data no matter the file type


client/src/GettingStarted/GettingStarted.js
```diff
@@ -32,7 +32,7 @@ export function GettingStarted() {
     }
 
     getUser()
-  }, [])
+  }, [makeRequest])
```
arku, line 35: @katelovescode This causes infinite requests to be sent to `/api/v1/profile`. You can observe it by inspecting the Network tab.

katelovescode: Yeah, I saw this, but when I removed the dependency it didn't fix it. O_O Maybe I'm missing something.  Will look again tomorrow.

katelovescode: I was mistaken; apparently it worked this morning 🤷

client/src/CasesImport/CasesImport.js
```diff
@@ -14,48 +13,62 @@ export function CasesImport() {
 
   const columns = getColumns(t)
 
+  const getType = file => {
+    const nameArray = file.name.split('.')
+    const extension = nameArray[nameArray.length - 1]
+    // sometimes xls files come in with an empty type so we want to parse
+    // the extension to verify what parser to use
+    if (file.type === '' && (extension === 'xls' || 'xlsx')) {
```
arku, line 21: I think you meant to write `extension === 'xls' || extension === 'xlsx`.

We could also write it as `['xls', 'xlsx'].includes(extension)`.

client/src/CasesImport/CasesImport.js
```diff
@@ -14,48 +13,62 @@ export function CasesImport() {
...
     setError('')
 
     const [file] = e.target.files
-    const { type } = file
-    if (!(type in parserTypes)) {
-      setKids([])
-      setError(t('uploadCasesInvalidFileFormat'))
-      return
-    }
+    if (file) {
```
arku, line 32: A minor thing. AFAIK, `file` will never we falsy since `onFileChange` is only called when a file is uploaded.

client/src/GettingStarted/GettingStarted.js
```diff
@@ -32,6 +32,8 @@ export function GettingStarted() {
     }
 
     getUser()
+    // we only want this to run once; making the makeRequest hook a dependency causes an infinite re-run of this query
```
arku, line 35: 👌

client/src/CasesImport/CasesImport.js
katelovescode: I sure did mean to write that.  Will fix.

client/src/CasesImport/CasesImport.js
katelovescode: Will fix

katelovescode: I actually think it is empty if you cancel out of the file menu; just remembered I did that on purpose last night.

arku: Ah, interesting.

arku: I just tried to see if the `onChange` event is triggered when the file explorer is closed without selecting a file but it wasn't. Not sure if I am missing something 😕

arku (dismissed):
Just tested it locally and LGTM 👍 

Thanks for fixing the type detection and adding E2E tests 😍

PR #293, opened 2020-09-06
Add ChildCaseCycle model

arku:

#160

Adds the migrations, model, factory and specs.

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally
Run `rails db:migrate`


spec/factories/child_case_cycles.rb
```diff
@@ -0,0 +1,41 @@
+FactoryBot.define do
+  factory :child_case_cycle do
+    user
+    child
+    subsidy_rule
+    case_cycle
+    part_days_allowed { 1 }
+    full_days_allowed { 1 }
```
arku, line 8: I wasn't sure of the reasonable defaults for these attributes. Let me know what values we should use here.

weedySeaDragon (approved):
1 minor thing to change:  the `dbdiagram.dbml` needs to include the relationship to a User:
```

Table child_case_cycles {
  id uuid [pk, unique, not null]
  user_id uuid [ref: > users.id, not null]
  ...
```

And still assuming we'll come thru later issues/PRs for any controller stuff, `seeds.db` and the api/swagger stuff.  (Doing it this way makes total sense to me.  Kate can make sure the basic model is correct, then we can do the other stuff.)

weedySeaDragon (approved):

spec/factories/child_case_cycles.rb
katelovescode: I'd do a random number between 10 and 20 maybe

katelovescode (approved):

Issue #166, opened 2020-08-02
DB Implementation: Child Case Cycle Payments


katelovescode:
## 📋 Summary
Write the migration, model and controller actions necessary for basic CRUD on Child Case Cycle Payments from the data model.

## 💻 Implementation Details
- This is an authenticated route
- Users will be associating child case cycles to payments on their account and will need authorization (i.e. they can only add/edit/update/delete child case cycle payments associated to their account)
- All CRUD actions are required
- see docs/dbdiagram.dbml for field definitions and associations
- This is blocked by #160 and #165  

## 🧐 Why tho
Start getting the DB built to make front end tickets easier

PR #319, opened 2020-09-08
add ChildCaseCyclePayments model, controller, api, swagger info

weedySeaDragon:

#166 

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🕯 Caveats, concerns

1. a number of models set their `slug` based on the `id`.  I created a protected method in ApplicationRecord that child classes can call: `set_slug_from_id`.  ChildCaseCyclePayment uses it: `before_save :set_slug_from_id`

2. For all of the factories, I replaced all user association with a `confirmed_user` so that the **confirmation email is not created.**  (It slows things down and fills up the test.log).
It's clear if you look at the factories.  I couldn't think of a reason not to use a confirmed_user.


app/models/application_record.rb
```diff
@@ -13,4 +13,10 @@ def generate_slug(hash_string)
   rescue ActiveRecord::RecordNotUnique
     Digest::SHA1.hexdigest("#{hash_string}#{SecureRandom.uuid}")[8..14] if hash_string
   end
+
+  protected
+
+  def set_slug_from_id
```
katelovescode, line 19: We're actually going to be getting rid of slugs in a later ticket, per our discussion in Slack - I don't think you need to change this now but we're gonna end up pulling it out. @csprayregen can we prioritize the slug ticket soonish?

spec/factories/businesses.rb
```diff
@@ -4,7 +4,7 @@
   factory :business do
     name { Faker::Name.child_care_businesses }
     license_type { Licenses.types.keys.sample }
-    user
+    user factory: :confirmed_user
```
katelovescode, line 7: 👍🏼😍👍🏼😍👍🏼😍

katelovescode (approved):
Looks great!

app/models/application_record.rb
weedySeaDragon: just saw that.  Yup -- we can decide which way to go (uuids v. slugs) and go from there.

Issue #163, opened 2020-08-02
DB Implementation: Attendances


katelovescode:
## 📋 Summary
Write the migration, model and controller actions necessary for basic CRUD on Attendances from the data model.

## 💻 Implementation Details
- This is an authenticated route
- Users will be managing attendances on their account and will need authorization (i.e. they can only add/edit/update/delete attendances associated to their account)
- All CRUD actions are required
- see docs/dbdiagram.dbml for field definitions and associations
- this is blocked by #160 

## 🧐 Why tho
Start getting the DB built to make front end tickets easier

PR #320, opened 2020-09-08
add Attendance model, controller, api, swagger; DateValidation invalid_date_msg 

weedySeaDragon:

#163 

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🕯 Caveats, concerns

I added a public method to DateParamValidator to access the _invalid date message_ so it can be used in RSpecs.


app/models/attendance.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# Attendance of a child during a specific cycle for a child case
+class Attendance < UuidApplicationRecord
+  belongs_to :child_case_cycle
+
+  LENGTHS_OF_CARE = %w[part_day full_day full_plus_part_day full_plus_full_day].freeze
```
katelovescode, line 7: @csprayregen when someone adds an attendance, are we expecting them to choose what length of care it is, or are we calculating that by the number of hours as laid out in their subsidy rule? 

If we're calculating it, can the user change it, if for some reason our calculation gets it wrong?

katelovescode (commented):
Outstanding business logic question for Chelsea

app/models/attendance.rb
csprayregen: Since we haven't designed these screens yet, I don't think we have clear answers. My assumption, though, is that we're calculating it for them based on the number of hours, as laid out in their subsidy rule. I'm not sure if we'll let them change it.

weedySeaDragon: I'll put a note in the model so these comments are captured.

app/models/attendance.rb
```diff
@@ -0,0 +1,57 @@
...
+#
+# from GitHub comments on the PR to implement this model (PR 320)
+# 2020-09-08: open question from Kate D.:
+#   When someone adds an attendance, are we expecting them to choose what
+#   length of care it is, or are we calculating that by the number of hours
+#   as laid out in their subsidy rule?
+#   If we're calculating it, can the user change it, if for some reason our
+#   calculation gets it wrong?
+#
+#   2020-09-08: Answer from Casey S:
```
katelovescode, line 13:
```suggestion
#   2020-09-08: Answer from Chelsea S:
```

weedySeaDragon: whoops

app/controllers/api/v1/attendances_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  private
+
+  def set_attendance
+    @attendance = Attendance.find_by!(slug: params[:slug])
+  end
+
+  # Only allow a trusted parameter "white list" through.
+  def attendance_params
+    params.require(:attendance).permit(
+      :id, :child_case_cycle_id, :slug, :length_of_care, :starts_on
```
katelovescode, line 53: Per the conversation with Chelsea, can we remove length of care from the allowed params? And maybe put a TODO  in new/create to calculate length of care based on subsidy rules?

weedySeaDragon: will do.

katelovescode (dismissed):

db/migrate/20200907004651_create_attendances.rb
```diff
@@ -0,0 +1,12 @@
+class CreateAttendances < ActiveRecord::Migration[6.0]
+  def change
+    create_table :attendances, id: :uuid do |t|
+      t.references :child_case_cycle, type: :uuid, null: false, foreign_key: true
+
```
katelovescode, line 5: We missed two fields in the data model; `check_in_time` and `check_out_time` - waiting for confirmation from Chelsea if that's the right name for the fields, but they should both be `TIME` type fields, formatted only to the minute.  We have some open questions so I think we should keep this PR on hold until we get them answered.

Open questions:

- Is check_in_time and check_out_time the right way to do it?
- Also are those times rounded for calculation of length of care?  Or are they to-the-minute?

katelovescode: This model should have `starts_on`, `check_in`, `check_out`, `total_time_in_care`, `slug`, and let's rename `length_of_care` to `attendance_duration`.  We'll do the rounding/logic in a later ticket.

katelovescode: Interesting:  https://www.postgresqltutorial.com/postgresql-interval/  I've never used this data type before.  This might suit our needs for `total_time_in_care`?

weedySeaDragon: Cool.  postgres `interval` + Ruby `Duration` = yes, that looks like it'd work!

weedySeaDragon:
@katelovescode :  I made the changes to Attendance as you asked.
I also added in a relation to `child_site` because that seems like it _really_ needs it. 

Annnd  I added in a bunch of code into `db/seeds.rb` so that Attendance records are created and seeded.   There are lots of ways to organize code used to seed things (make some 'helper files' to just keep the code better organized, etc etc.  I left this as one big honkin' file for now; we (you) can determine how you want to deal with it.  

I wanted to just go ahead and get this in so that we can then start noodling/percolating on modeling the _lengths_of_care_ and _subsidy_rules_ and _definitions of things_  etc.

app/models/attendance.rb
```diff
@@ -17,41 +17,58 @@
 #   I'm not sure if we'll let them change it.
 #
 class Attendance < UuidApplicationRecord
-  belongs_to :child_case_cycle
+  belongs_to :child_site
+  has_one :child, through: :child_site
+  has_one :site, through: :child_site
+
+  belongs_to :child_case_cycle # TODO: really?
```
katelovescode, line 24: I think so; we could do the calculation based on dates but I think getting the indexing to its child_case_cycle for free might help with performance later.

app/validators/time_param_validator.rb
```diff
@@ -0,0 +1,16 @@
...
+  INVALID_TIME_MSG = 'Invalid time'
+
+  def self.invalid_time_msg
+    INVALID_TIME_MSG
+  end
+
+  def validate_each(record, attribute, value)
+    value&.is_a?(Time) ? value : Time.zone.parse(value)
+  rescue TypeError, ArgumentError
+    record.errors.add(attribute, self.class.invalid_time_msg)
```
katelovescode, line 14: ❤️

app/models/attendance.rb
```diff
@@ -17,41 +17,58 @@
...
 
   LENGTHS_OF_CARE = %w[part_day full_day full_plus_part_day full_plus_full_day].freeze
```
katelovescode, line 26: Rename to `DURATION_DEFINITIONS`?

db/migrate/20200913004651_create_attendances.rb
```diff
@@ -1,10 +1,14 @@
...
     create_table :attendances, id: :uuid do |t|
+      t.references :child_site, type: :uuid, null: false, foreign_key: true
       t.references :child_case_cycle, type: :uuid, null: false, foreign_key: true
-
       t.string :slug, null: false, index: { unique: true }
+
       t.date :starts_on, null: false
+      t.time :check_in, null: false
+      t.time :check_out, null: false
+      t.interval :total_time_in_care, null: false, comment: 'Calculated: check_out time - check_in time'
```
katelovescode, line 11: 😱 😍

app/models/attendance.rb
```diff
@@ -17,41 +17,58 @@
...
+  validates :check_out, time_param: true
 
   private
 
   def set_slug
     self.slug = generate_slug("#{SecureRandom.hex}#{id}")
   end
+
+  def calc_total_time_in_care
+    self.total_time_in_care = (check_out.nil? || check_in.nil? ? 0 : check_out - check_in)
```
katelovescode, line 44: 👍

db/seeds.rb
```diff
@@ -22,13 +37,46 @@
...
+puts_records_in_db(User)
+
+# ---------------------------------------------
+# Children
+# ---------------------------------------------
+
+
+# find_or_create_by! a Child with the full_name,
+#  and birthday set randomly between the min_age and max_age.
+def child_named(full_name, min_birthday: (Time.zone.now - 2.weeks),
```
katelovescode, line 50: Can we extract these min_birthday and max_birthday into constants above?  These might come in handy later.

spec/models/child_site_spec.rb
```diff
@@ -5,6 +5,8 @@
 RSpec.describe ChildSite, type: :model do
   it { should belong_to(:child) }
   it { should belong_to(:site) }
+
+  # TODO: validate started_care, ended_care
```
katelovescode, line 9: I don't think this will be required for now, so the only validations I can think of are date validations

katelovescode (commented):
Looks great, just a couple comments/requests

app/models/attendance.rb
weedySeaDragon: will do

katelovescode (approved):

Issue #318, opened 2020-09-08
After I verify my account, the "resent verification email" button should produce an error


csprayregen:
## ✨ What new feature do you want to see?
After I create an account, but before I verify the account, I can re-send the account verification email on the "thanks for signing up page." If I still have that page open, but I've already verified my account, I do not get an email, but I also don't get an error message. I'd like to get an error message in this instance to make this clear to users. 

## 🧶 Acceptance Criteria
- [x] After my account is verified, if I click "click here to resend the email", I get an error message that says, "You have already verified your account. You can now log in." 

- [x] If I've selected Spanish, this text should read, "Ya has verificado tu cuenta. Ahora puedes iniciar sesión."

## 🖍️ Implementation Details
@katelovescode to add any technical implementation details I might have missed. 

## 🥧 Additional context
We covered the main functionality of this feature in #195, but this is an additional item.

PR #321, opened 2020-09-08
Display an error message if the user has already confirmed their email address

arku:

https://github.com/pieforproviders/pieforproviders/issues/318

If the user clicks on the **Click here to resend the email** link after confirming their email address, they are redirected to the login page and an error message is shown.

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test

1. Sign up for an account. You should be taken to the confirmation instructions page
2. Click on the link in the confirmation email sent to your email address
3. Go back to the confirmation instructions page and click the **Click here to resend the email** link. You should be taken to the login page and the error message _You have already verified your account. You can now log in._ should show up

Note that the alert title **Confirmation Error** will not be translated upon switching the language to Spanish. This is because the message comes from the backend. The fix involves updating the code which displays login errors and will be handled in a separate PR.

### DIFF
--- a/client/src/i18n/locales/en/index.json
+++ b/client/src/i18n/locales/en/index.json
@@ -84,7 +84,7 @@
 
   "genericErrorMessage": "Something went wrong, try again later.",
 
-  "alreadyConfirmed": "This email has already been confirmed, please try logging in. If you've forgotten your password, please reset it below.",
+  "alreadyConfirmed": "You have already verified your account. You can now log in.",
   "emailUnconfirmed": "You have to confirm your email address before continuing.",
   "confirmationPeriodExpired": "Your confirmation period has expired.",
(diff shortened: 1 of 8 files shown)


client/src/i18n/locales/en/index.json
```diff
@@ -77,7 +77,7 @@
 
   "genericErrorMessage": "Something went wrong, try again later.",
 
-  "alreadyConfirmed": "This email has already been confirmed, please try logging in. If you've forgotten your password, please reset it below.",
+  "alreadyConfirmed": "You have already verified your account. You can now log in.",
```
katelovescode, line 80: Did we get this language from some place?

katelovescode (dismissed):

katelovescode (approved):

Issue #178, opened 2020-08-10
Add Pundit for user authorization


katelovescode:
## 📋 Summary
Restrict user access to their records only (as a test case, users can only see their own businesses)

## 💻 Implementation Details
Here's a place to start looking: https://github.com/stalniy/rails-cancan-api-example

## 🧐 Why tho
We don't want users to be able to see records that don't belong to them!  We can help to enforce it with API endpoint authorization rules.

katelovescode:
@csprayregen for prioritization

PR #331, opened 2020-09-10
Set up Pundit

arku:

#178

- Adds an admin column to the users' table
- Sets up Pundit 
- Adds policies for users and businesses

I'll open separate PRs with policies for the rest to keep this one small.

## 🍂 Before You Submit

* [x] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [x] Dependencies

## 🧵 Steps to set up locally

Run `bundle`

## 🕯 Caveats, concerns

Questions:

When a user tries to access a business or a site that isn't theirs, a 404 is returned. This is what Github does when a visitor tries to access a private repository. Is this fine?


app/policies/application_policy.rb
```diff
@@ -0,0 +1,41 @@
...
+  class Scope
+    attr_reader :user, :scope
+
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
+      scope.all
```
weedySeaDragon, line 38: We should remove this an define the Scopes on a _per-class basis._

(If we did want to put it here for some reason, then `resolve` should be:
 if the user is an admin, the scope is `all`
 else the scope is none.  (return an empty list)
)

app/policies/application_policy.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+# Base policy class
+class ApplicationPolicy
+  attr_reader :user, :record
```
weedySeaDragon, line 5: An `admin` should have permission to see and do any actions.
The default result for _action query methods_ (`index? create?` etc) should all call this method:

```
def allowed_default_access?
   user.admin?
end
```

So the action query methods (`index? create?` etc)  should

weedySeaDragon (changes requested):

app/policies/application_policy.rb
arku: I've only retained the constructor in the base scope class.

app/policies/application_policy.rb
arku: > The default result for action query methods (index? create? etc) should all call this method:

Any authenticated user can create resources. Except `UserController#index` which only admins can access, all other controllers' index actions are accessible by authenticated users — what resources they get to access is defined by the policy scope.

I have used the `user.admin?` predicate as the default for `update` and `destroy` actions as you suggested.

swagger/v1/swagger.json
```diff
@@ -155,11 +155,6 @@
               "schema": {
                 "$ref": "#/components/schemas/updateBusiness"
               }
-            },
-            "application/xml": {
```
arku, line 159: Let me know if we should add thie back.

app/policies/business_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+
+# Authorization policies for businesses
+class BusinessPolicy < ApplicationPolicy
+  # Scope defining which businesses a user has access to
+  class Scope < Scope
+    def resolve
+      if user.admin?
+        scope.all
+      else
+        scope.where(user: user).active
```
katelovescode, line 11: 😍

app/policies/business_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  private
+
+  def manage?
+    admin? || owner?
+  end
+
+  def owner?
+    record.user == user
```
katelovescode, line 31: 😍

db/structure.sql
```diff
@@ -61,8 +61,6 @@ CREATE TYPE public.license_types AS ENUM (
 
 SET default_tablespace = '';
 
-SET default_with_oids = false;
```
katelovescode, line 64: 😠  whyyyyyy

swagger/v1/swagger.json
katelovescode: 👎  don't think we need this

katelovescode (approved):

katelovescode:
When conflicts are resolved, I'll re-approve and merge

app/controllers/api/v1/businesses_controller.rb
```diff
@@ -29,6 +33,7 @@ def create
 
   # PATCH/PUT /businesses/:slug
   def update
+    authorize @business
```
weedySeaDragon, line 36: instead use  `before_action :authorize_object,  only: %i[update destroy]`

and then define the private method:

```
def authorize_business
    authorize @business
end
```

app/controllers/api/v1/users_controller.rb
```diff
@@ -30,6 +31,8 @@ def create
 
   # PATCH/PUT /users/:slug
   def update
+    authorize @user
```
weedySeaDragon, line 34: instead use  `before_action :authorize_object,  only: %i[update destroy]`

and then define the private method:

```
def authorize_user
    authorize @user
end
```

app/controllers/api/v1/businesses_controller.rb
```diff
@@ -38,19 +43,20 @@ def update
 
   # DELETE /businesses/:slug
   def destroy
+    authorize @business
```
weedySeaDragon, line 46: see my comment about authorization in `update`

app/controllers/api/v1/users_controller.rb
```diff
@@ -39,6 +42,7 @@ def update
 
   # DELETE /users/:slug
   def destroy
+    authorize @user
```
weedySeaDragon, line 45: see my comment about authorization in `update`

app/policies/business_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+    manage?
+  end
+
+  private
+
+  def manage?
+    admin? || owner?
+  end
+
+  def owner?
```
weedySeaDragon, line 30: This is a common need, so put it in the parent class. (as `protected`)

app/policies/business_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+    manage?
+  end
+
+  def destroy?
+    manage?
+  end
+
+  private
+
+  def manage?
```
weedySeaDragon, line 26: This is a common need, so put it in the parent class. (as `protected`)
Also: "manage" does not convey anything about what this is checking/querying.  ('manage' is vague; it also could have different meanings in different contexts)
These behaviors/queries/methods are all about _who has permissions_  so the method should be named `admin_or_owner?`

app/policies/user_policy.rb
```diff
@@ -0,0 +1,37 @@
...
+  def update?
+    manage?
+  end
+
+  def destroy?
+    manage?
+  end
+
+  private
+
```
weedySeaDragon, line 29: see my comments in the BusinessPolicy class  re `manage?` and `owner`

app/policies/application_policy.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
+
```
weedySeaDragon, line 11: I think that the most common permissions are (and thus should be the defaults in this parent class):
```

  def index?
    admin_or_owner?
  end

  def create?
    true
  end

  def update?
    admin_or_owner?
  end

  def destroy?
    admin_or_owner?
  end
```

Then the `UserPolicy` only needs to override these:
```
class UserPolicy
  ...
  def index?
    admin?
  end


  # The user is looking at their own record; they 'own' their own record
  def owner?
    user == record
  end

```

Then `BusinessPolicy`  and most (if not all) other Policy classes only need to define their Scope `resolve` method

app/policies/business_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+    def resolve
+      if user.admin?
+        scope.all
+      else
+        scope.where(user: user).active
+      end
+    end
+  end
+
+  def update?
```
weedySeaDragon, line 16: not needed; see my comment in ApplicationPolicy

app/policies/business_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+    manage?
+  end
+
+  def destroy?
```
weedySeaDragon, line 20: not needed; see my comment in ApplicationPolicy

app/policies/user_policy.rb
```diff
@@ -0,0 +1,37 @@
...
+        scope.where(id: user.id).active
+      end
+    end
+  end
+
+  def index?
+    admin?
+  end
+
+  def update?
```
weedySeaDragon, line 20: not needed; see my comment in ApplicationPolicy

app/policies/user_policy.rb
```diff
@@ -0,0 +1,37 @@
...
+    manage?
+  end
+
+  def destroy?
```
weedySeaDragon, line 24: not needed; see my comment in ApplicationPolicy

app/policies/application_policy.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+# Base policy class
+class ApplicationPolicy
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    @user = user
```
weedySeaDragon, line 8: This guard should be the first line in the method:
```
# TODO: use I18n.t() for 'must be logged in' message
    raise Pundit::NotAuthorizedError, "must be logged in" unless user
```

app/policies/application_policy.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def destroy?
+    admin?
+  end
+
+  # Base policy scope class
+  class Scope
+    def initialize(user, scope)
+      @user  = user
```
weedySeaDragon, line 31: This guard should be the first line in the method:
```
# TODO: use I18n.t() for 'must be logged in' message
    raise Pundit::NotAuthorizedError, "must be logged in" unless user
```

weedySeaDragon (changes requested):
(1) We can simplify/DRY a _lot_ because we know that most policies will use the same permissions for the common REST actions.  See my comments in the files.

(2) We can also simplify/DRY up the `Scope` `resolve` methods for each Policy. Seems there are 3 pieces that can be chained together as needed: a user can only get 'their' objects; an admin can get all;  a user should only get the 'active' objects they own. Once we implement more of the Policies and the `Scope resolve`  methods, the patterns will be clear.  

(3) We can also generalize/abstract the logic in the `.permit(attributes)...` stuff.  If a class has an `active` attribute, then they're going to have the ability to have that scope applied for `Policy .. Scope resolve` method, _and_ have it be a permitted attribute only for an admin.
That means that behavior can be abstracted out into a module.

(1) above should be done now.  But (2) and (3) can wait for a future PR. By then the patterns will be really clear.

db/structure.sql
arku: This shouldn't have been there in the first place. Looks like it was added again in https://github.com/pieforproviders/pieforproviders/pull/319 after being removed in https://github.com/pieforproviders/pieforproviders/pull/313.

I am using Postgres 12.4 and I believe you're as well. @weedySeaDragon Can you check what version of Postgres you're running?

app/controllers/api/v1/businesses_controller.rb
arku: I named it `authorize_user` since it is the user we're authorizing.

app/policies/application_policy.rb
arku: I didn't add it as we already make sure [only authenticated users can access the API](https://github.com/pieforproviders/pieforproviders/blob/develop/app/controllers/api/v1/api_controller.rb#L5). But it doesn't hurt adding the guard I guess.

arku: We don't have to bother translating the message since users will never get to see it.

app/models/business.rb
```diff
@@ -10,6 +10,8 @@ class Business < UuidApplicationRecord
   validates :name, presence: true, uniqueness: { scope: :user_id }
 
   before_validation { |business| business.slug = generate_slug("#{business.name}#{business.user_id}") }
+
+  scope :active, -> { where(active: true) }
```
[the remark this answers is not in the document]
katelovescode: Scope creep, get thee out of here. :D Drop it waaaay down in the backlog.

katelovescode (approved):

Issue #269, opened 2020-09-01
DB implementation: Connect Address lookups to models


weedySeaDragon:
## 📋 Summary

Connect the Address lookup models (`State` `County` `City`) to the models that use them.

blocked by #338 

## 💻 Implementation Details

Will need schema migrations, model and spec updates.

## 🧐 Why tho

Now that we have the address lookup models in the db, we should use them.

PR #339, opened 2020-09-12
Connect Sites, Agencies to Lookup address classes (based on PR 336)

weedySeaDragon:

#269 Connect Address lookups to models

This branch is based on the branch for PR 336.  Once that PR is ok and merged, I'll update this as necessary.

## 🍂 Before You Submit

* [ ] Did you write tests?

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🕯 Caveats, concerns

I also added an explicit example (RSpec test) to models to verify that the default factory is valid.


db/seeds.rb
```diff
@@ -28,44 +28,56 @@
   license_type: Licenses.types.keys.first
 )
 
-site = Site.where(name: 'Prairie Center', business: business).first_or_create(
+montana = Lookup::State.find_by(abbr: 'MT')
```
katelovescode, line 31: I know we're expecting these to be populated by the import but should these be `first_or_create`s just in case?

katelovescode (dismissed):
Just a comment, not a dealbreaker, merge if you are satisfied!

katelovescode (approved):

Issue #333, opened 2020-09-11
"Email taken" warning on sign up page doesn't disappear


csprayregen:
## 🤬 What's wrong?
When I'm on the sign up page, and I input an email address that has already been taken, the warning stays there, even after I fix this problem. The user would not be sure she had fixed the error in the current state. 

## 👉🏼 Steps to Reproduce

<!-- Steps to reproduce the behavior:
1. Go to https://pie-staging.herokuapp.com/signup
2. Go through the steps to create an account, but use an email address that has already been taken. 
3. You'll see the warning message appropriately appear stating "email address is already taken."
4. Update the email to a new address.
5. The warning does not go away. However, I can successfully create a new account once I input a new email address. 

## 🧥Expected behavior
The warning should go away once I input a new email address, so that I know I can create an account.

## 📸 Screenshots
![Screen Shot 2020-09-11 at 11 20 28 AM](https://user-images.githubusercontent.com/26717304/92949964-8b83f300-f421-11ea-87e5-d5611749ef46.png)

## 💻 Desktop

Please complete the following information if accessing the app from a computer:

- OS: Catalina 10.15.4
- Browser: Chrome
- Version [e.g. 22]:  85.0.4183.83 (Official Build) (64-bit)

## 📱Smartphone

n/a

csprayregen:
@arku @katelovescode this does not appear to be fixed. See that the error remains in this example

![Screen Shot 2020-09-18 at 9 36 26 AM](https://user-images.githubusercontent.com/26717304/93610126-4cf6b700-f992-11ea-9b1e-328d25d24f55.png)

Additionally, this error does not appear at all when there are other errors. It seems to take this one as last priority. 
![Screen Shot 2020-09-18 at 9 34 45 AM](https://user-images.githubusercontent.com/26717304/93610167-57b14c00-f992-11ea-9064-53da42eb5709.png)

naresh3599:
Hii @csprayregen , would like to work on this, 
please assign me!

csprayregen:
Done @naresh3599 !

narendran-kannan:
I need quite more details on how to reproduce this @csprayregen 
Do we need to remove that error msg "Email has already been taken" after they've changed their email in frontend?
Correct me if I'm wrong

katelovescode:
> I need quite more details on how to reproduce this @csprayregen
> Do we need to remove that error msg "Email has already been taken" after they've changed their email in frontend?
[...]

PR #354, opened 2020-09-15
Move signup validation error translations to the frontend

arku:

https://github.com/pieforproviders/pieforproviders/issues/333

- Updates the error response to include the [error details](https://guides.rubyonrails.org/active_record_validations.html#validations-overview-errors-details) with the validator names (e.g. blank, taken, etc)

- Defines translations for the validator names.

## 🍂 Before You Submit

* [ ] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test

1. Sign up with an email address that already exists in the database
2. You should find the message "Email has already been taken"  (if the selected language is English)
3. Switch the language to Spanish and you should find the message "Correo ya ha sido tomado"
4. Sign up with a phone number that already exists and the error message should be translatable

### DIFF
--- a/client/src/Signup/Signup.js
+++ b/client/src/Signup/Signup.js
@@ -53,7 +53,7 @@ export function Signup() {
       setSuccess(true)
     } else if (response.status === 422) {
       const { errors } = await response.json()
-      setValidationErrors(errors[0].detail)
+      setValidationErrors(errors)
     } else {
       setError(true)
     }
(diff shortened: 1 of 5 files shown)


client/src/Signup/Signup.js
```diff
@@ -53,7 +53,7 @@ export function Signup() {
       setSuccess(true)
     } else if (response.status === 422) {
       const { errors } = await response.json()
-      setValidationErrors(errors[0].detail)
+      setValidationErrors(errors)
```
katelovescode, line 56: 😍

katelovescode (approved):

Issue #180, opened 2020-08-11
As a user, I expect not to see internal pages after my token has expired


katelovescode:
## ✨ What new feature do you want to see?
Currently, if I visit an "internal" page (getting-started for example) with any token in my localStorage, it will show me the page; presumably any API calls would fail, but the non-api content still loads.  I'd expect that if my token is expired, I will be redirected to the login page to log in again

## 🧶 Acceptance Criteria
- If it has been more than 24 hours since I logged in, when I visit an authorized route (i.e. 'getting-started'), I am redirected to the login page.

## 🖍️ Implementation Details
Tokens expire by default after 1 day, currently.  For development that can be changed at `config/initializers/devise.rb:309`.  The best implementation I've come up with is to store the expiration time in localStorage as well, and then check it on the front end around `client/src/_utils/_routes/AuthorizedRoute.js:15`, and redirect if it's expired.  Would love to hear other implementations if there are any.

New note re decoding client-side: https://github.com/pieforproviders/pieforproviders/pull/377#discussion_r494703168

## 🔮 Testing instructions

- [x]  Log in
- [x]  In your JS console, do: localStorage.getItem('pie-token') and localStorage.getItem('pie-expiration'), they should be set
![Screen Shot 2020-10-18 at 7 45 49 PM](https://user-images.githubusercontent.com/8364647/96390482-8c274c00-117a-11eb-8601-92ce988d77c3.png)
- [x]  Log out
- [x]  In your JS console, do: localStorage.getItem('pie-token') and localStorage.getItem('pie-expiration'), they should be null
![Screen Shot 2020-10-18 at 7 46 45 PM](https://user-images.githubusercontent.com/8364647/96390517-ab25de00-117a-11eb-8c0b-59b0b9f6da96.png)
- [x]  Log in
- [x]  In your JS console, do localStorage.setItem('pie-expiration', '[one day earlier than the initial setting]') <= i.e. change the Date string from today's date to yesterday's date
![Screen Shot 2020-10-18 at 7 48 31 PM](https://user-images.githubusercontent.com/8364647/96390572-eaecc580-117a-11eb-94c2-d6a80c49883f.png)
- [x]  Visit /dashboard <= this is a page that doesn't make any db calls, but is an authenticated route
- [x]  You should be redirected to /login
- [x]  Try to visit /dashboard again, you should be redirected to /login again

katelovescode:
@csprayregen ready for prioritization

PR #365, opened 2020-09-20
[180] Auth Context

katelovescode:
Related to/came out of #180 but doesn't finish the actual point of that ticket.  This just moves localStorage calls to a use-local-storage-state hook, and moves all the auth functionality into a context to be shared with other components.

## 🍂 Before You Submit

* [ ] Did you write tests?

* [ ] Data Migrations
* [ ] Schema Migrations
* [X] Dependencies

## 🧵 Steps to set up locally

`yarn --cwd client install`

## 🧳 Steps to test
Check all the authorized routes and non-authorized routes and make sure it's working as expected

## 🕯 Caveats, concerns
Still trying to figure out the expiration stuff; this has expiration in the context, but it's not used, it's a placeholder for expiration behavior if we can get it to work.


client/src/i18n/locales/es/index.json
```diff
@@ -63,6 +63,7 @@
   "resendConfirmationEmail": "Haga clic aquí para reenviar el correo.",
   "toResendConfirmationEmail": "para reenviar el correo de confirmación.",
   "confirmationEmailResent": "Correo reenviado",
+  "confirmYourAccount": "Confirme tu cuenta",
```
katelovescode, line 66: @csprayregen please confirm translation

client/src/_utils/_routes/AuthenticatedRoute.js
```diff
@@ -0,0 +1,39 @@
...
+  path,
+  title,
+  // permissions,
+  ...routeProps
+}) {
+  exact = !!exact
+  let history = useHistory()
+  const authenticated = IsAuthenticated
+
+  useEffect(() => {
```
DoodlingDev, line 20: I'm curious about this `useEffect` here. Without a dependency array, does it run every time the function is re-rendered? So it's the same as not being in a use Effect?

katelovescode: Good point, it's probably not necessary to do it in the useEffect, gonna test that theory.

katelovescode:
Closing this in favor of redux

PR #377, opened 2020-09-22
Move Local Storage behavior to hooks

katelovescode:
N/A

## 🍂 Before You Submit

* [ ] Did you write tests?

## 🧳 Steps to test
Visit the auth'ed and non-auth'ed routes and make sure they behave as expected.  Log in, log out, reset password, etc.

### DIFF
--- a/client/src/_shared/_hooks/useAuthentication.js
+++ b/client/src/_shared/_hooks/useAuthentication.js
@@ -8,7 +8,7 @@ export function useAuthentication() {
 
   const isAuthenticated =
     !!localStorage.getItem('pie-token') &&
-    localStorage.getItem('pie-expiration') > dayjs()
+    dayjs(localStorage.getItem('pie-expiration')).isAfter(dayjs())
 
   const removeToken = () => {
     localStorage.removeItem('pie-token')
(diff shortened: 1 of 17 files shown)


client/src/_shared/_hooks/useAuthentication.js
```diff
@@ -0,0 +1,24 @@
+import dayjs from 'dayjs'
+
+export function useAuthentication() {
+  const setToken = (token, expiration = dayjs().add('1', 'day')) => {
+    localStorage.setItem('pie-token', token)
+    localStorage.setItem('pie-expiration', expiration)
```
arku, line 6: **question**: Can't we decode the JWT client-side (we don't need the secret key because we're not validating the token) using [jwt-decode](https://github.com/auth0/jwt-decode)  and extract expiration `exp` from the token?

arku (approved):
LGTM 👍 

Left a comment to simplify the expiration management.

client/src/_shared/_hooks/useAuthentication.js
katelovescode: Oooooooooh interesting point.  I guess I hadn't thought of it that way.  I'm going to investigate!  Thank you!

PR #388, opened 2020-09-30
Rename AuthorizedRoute to AuthenticatedRoute

katelovescode:
Chore related to #180 - renaming route to be more semantic

## 🍂 Before You Submit

## 🧳 Steps to test
Visit an authenticated route while logged in, it should work
Visit an authenticated route while logged out, it should stop working


arku (approved):
LGTM 💯

PR #409, opened 2020-10-06
[180] move auth info to global state redux, expiration handling on the front end

katelovescode:
#180 for real!

This implements Redux as our global state handler and moves authentication logic into reducers rather than custom hooks.  This also adds decoding the JWT for expiration date, storing it in state (and localStorage for persisted logins), and checking for expired tokens before displaying authenticated routes.  The purpose of this is that someone might go through onboarding (for example) and enter a lot of data that isn't being sent to the backend, then try to hit the backend and their token has expired, and they're logged out.  This is really preventative; when we had more complicated onboarding this was important but right now it's not as likely.

## 🍂 Before You Submit

* [X] Did you write tests?

* [X] Dependencies
  * `jwt-decode`

## 🧵 Steps to set up locally

`yarn install` in the `client` folder (or `yarn install-all` from the root)

## 🧳 Steps to test
- [ ] Log in
- [ ] In your JS console, do: `localStorage.getItem('pie-token')` and `localStorage.getItem('pie-expiration')`, they should be set
- [ ] Log out
- [ ] In your JS console, do: `localStorage.getItem('pie-token')` and `localStorage.getItem('pie-expiration')`, they should be null
- [ ] Log in
- [ ] In your JS console, do `localStorage.setItem('pie-expiration', '[one day later than the initial setting]')` <= i.e. change the Date string from today's date to tomorrow's date
- [ ] Visit `/dashboard` <= this is a page that doesn't make any db calls, but is an authenticated route
- [ ] You should be redirected to `/login`
- [ ] In your JS console, do: `localStorage.getItem('pie-token')` and `localStorage.getItem('pie-expiration')`, they should be null


client/src/_shared/_hooks/useAuthentication.js
```diff
@@ -1,24 +1,18 @@
 import dayjs from 'dayjs'
+import { useSelector } from 'react-redux'
 
+// TODO: Something isn't working here, if I hit an authenticated route directly, it logs me out
```
katelovescode, line 4: cc @arku @cjhaddad if you have a minute to look at this and see what I'm missing with grabbing the token from state when hitting an AuthenticatedRoute directly (i.e., login, land on getting-started, and then go to the browser bar and hit enter to hard reload getting-started and you'll be logged out)

client/src/Dashboard/Dashboard.js
```diff
@@ -1,13 +1,15 @@
...
 import { useTranslation } from 'react-i18next'
 import { useApiResponse } from '_shared/_hooks/useApiResponse'
-import { useAuthentication } from '_shared/_hooks/useAuthentication'
+import { useSelector } from 'react-redux'
 
 export function Dashboard() {
   const [businessList, setBusinessList] = useState([])
+  const token = useSelector(state => {
+    return state.auth.token
+  })
```
cjhaddad, line 10: nit: write this function like so: `state => state.auth.token`

client/src/GettingStarted/GettingStarted.js
```diff
@@ -7,26 +7,28 @@ import BusinessIcon from '@material-ui/icons/Business'
...
 // NB: we're using CSS grid instead of Ant grid for these cards
 // because Ant grid doesn't flow into the next row when there are
 // more cards than columns
 
 export function GettingStarted() {
   const [user, setUser] = useState(null)
+  const token = useSelector(state => {
+    return state.auth.token
+  })
   const { t } = useTranslation()
```
cjhaddad, line 22: same thing as comment in Dashboard

client/src/_actions/auth.js
```diff
@@ -0,0 +1,35 @@
...
+export const ADD_AUTH = 'ADD_AUTH'
+export const REMOVE_AUTH = 'REMOVE_AUTH'
+
+export function addAuth(token) {
+  const expiration = dayjs.unix(jwt_decode(token).exp).toDate()
+  /* 
+    we're using local storage here so we can re-sign the user in as long as
+    their token hasn't expired
+  */
+  localStorage.setItem('pie-token', token)
```
cjhaddad, line 13: NOICE

client/src/_actions/auth.js
```diff
@@ -0,0 +1,35 @@
...
+    expiration
+  }
+}
+
+export function removeAuth() {
+  const expiration = dayjs().toDate()
+  /* 
+    we're clearing local storage here on any logout event or anything that seems
+    like it should invalidate someone's "session"
+  */
```
cjhaddad, line 27: looovee these comments 💯

client/src/_actions/auth.js
```diff
@@ -0,0 +1,35 @@
+import dayjs from 'dayjs'
+import jwt_decode from 'jwt-decode'
+
+export const ADD_AUTH = 'ADD_AUTH'
+export const REMOVE_AUTH = 'REMOVE_AUTH'
+
+export function addAuth(token) {
+  const expiration = dayjs.unix(jwt_decode(token).exp).toDate()
```
cjhaddad, line 8: generally, these action methods are to be really, really simple and all the data/state changes should happen in the reducer. therefore, the logic for creating the expiration time, setting localStorage ,and setting the token to `null` done below should be relocated to the reducer.

cjhaddad (changes requested):

client/src/_reducers/auth.js
```diff
@@ -0,0 +1,20 @@
...
+import { ADD_AUTH, REMOVE_AUTH } from '_actions/auth'
+
+const initialAuthState = {
+  token: localStorage.getItem('pie-token'),
+  expiration: localStorage.getItem('pie-expiration') || dayjs().toDate()
+}
+
+export default function auth(state = initialAuthState, action) {
+  switch (action.type) {
+    case ADD_AUTH:
```
cjhaddad, line 11: follow up with my comment above: this case would be where you return the Object below. and the REMOVE_AUTH case would update the `token` to `null. 

related question for my own clarity -- expiration date is created upon adding the token, and then when removed, we keep it, or is it also removed? (to redux state, not localStorage)

client/src/_reducers/auth.js
```diff
@@ -0,0 +1,20 @@
...
+    case REMOVE_AUTH:
+      return Object.assign({}, state, {
```
cjhaddad, line 13: Instead of `Object.assign` I have been using the spread operator last couple years, which also allows you to create a new object via a merge of current state and updates. in this case it'd look something like this: {...state, token: action.token, expiration }

client/src/_reducers/auth.js
```diff
@@ -0,0 +1,45 @@
+import dayjs from 'dayjs'
+import jwt_decode from 'jwt-decode'
+import { ADD_AUTH, REMOVE_AUTH } from '_actions/auth'
+
+const initialAuthState = {
+  token: localStorage.getItem('pie-token'),
+  expiration: dayjs(localStorage.getItem('pie-expiration')) || dayjs()
```
arku, line 7: The`dayjs` function will always return an object — the second operand (`dayjs()`) will never get executed . Shouldn't it be the following? 🤔 

```
const token = localStorage.getItem('pie-token')
const expiration = localStorage.getItem('pie-expiration')
return {
  token,
  expiration: expiration ? dayjs(expiration) : dayjs()
}
```
```

File: client/package.json [hunk 10]
```diff
@@ -11,6 +11,7 @@
     "@craco/craco": "^5.7.0",
     "@material-ui/core": "^4.9.14",
     "@material-ui/icons": "^4.9.1",
+    "@reduxjs/toolkit": "^1.4.0",
```
arku, line 14: Looks like we're only using this package in tests. If we don't intend to use it elsewhere, can we install it as a dev dependency?

client/src/Login/Login.js
```diff
@@ -52,21 +53,23 @@ export function Login() {
     })
     if (!response.ok || response.headers.get('authorization') === null) {
       const errorMessage = await response.json()
+      dispatch(removeAuth())
```
arku, line 56: Do you mind explaining why we need this?

client/src/Login/Login.js
```diff
@@ -52,21 +53,23 @@ export function Login() {
...
       setApiError({
         status: response.status,
         message: errorMessage.error,
         attribute: errorMessage.attribute,
         type: errorMessage.type,
         context: { email: values.email }
       })
+      history.push('/login')
```
arku, line 64: We don't need this since the user is already on the login page.

client/src/_reducers/auth.js
katelovescode: Ooo good point.  I'll fix it.

client/package.json
katelovescode: Ooo good catch, but I should actually be USING the toolkit, I'll make a fix.

client/src/Login/Login.js
katelovescode: Whoops good catch

client/src/Login/Login.js
katelovescode: We probably don't.  I think I just put it in there as a proactive paranoid measure, clear out any stored data if they get a failed login attempt.  I'm okay with removing it.

client/src/_reducers/authReducer.js
```diff
@@ -0,0 +1,51 @@
...
+    token,
+    expiration: expiration || dayjs().format()
+  }
+}
+
+const auth = createSlice({
+  name: 'auth',
+  initialState: initialAuthState(),
+  reducers: {
+    addAuth(state, action) {
```
katelovescode, line 24: cc @cjhaddad check out this sweet Redux toolkit thing; createSlice lets you define reducers and actions in one place 💘   Gonna merge this because it's getting real stale, but we can change the implementation later if you want.

PR #465, opened 2020-10-21
[178] Pundit for attendances/billable occurrences, refactoring

katelovescode:

Finishes #178 with the new data model; we want Attendances to be created as instances of the polymorphic BillableOccurrence

I beefed up the request spec for displaying items in a user's scope; I create several records, only some of which belong to the current user, and make sure the number of displayed results is accurate.  I also refactored the requests specs because they were redefining several let blocks multiple times.

I changed all references to `item` in the request specs to `record` just because it has a little more meaning; I might even be convinced to call them `model` but I think that will cause confusion between requests specs and model specs.  I also added a shared example for updating nested items because of the new association between Attendances and BillableOccurrences.

## 🍂 Before You Submit

* [X] Did you write tests?

## 🧳 Steps to test

- run specs

- create several `BillableOccurrences` of type `Attendance` via the API (see the request spec for some ideas, and also the seeds file, where we make attendances by making `BillableOccurrences`) for several different users - the occurrences will need to be associated to some ChildApprovals, so make sure you create those required records first
- hit `/api/v1/billable_occurrences` and make sure:
	- you can only see the attendances for the current user
	- it includes the attendance data as well as the billable occurrence associated IDs

- create several business records as several different users
- hit `/api/v1/businesses` and make sure you can only see the current user's businesses


app/controllers/api/v1/billable_occurrences_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  private
+
+  def set_billable_occurrence
+    @billable_occurrence = BillableOccurrence.find(params[:id])
+  end
+
+  def billable_occurrence_params
+    params.require(:billable_occurrence).permit(:child_approval_id,
+                                                { attendance_params: %i[check_in check_out total_time_in_care] })
```
katelovescode, line 53: The current errors in my specs say:
```
12) billable occurrences API behaves like it updates an item /api/v1/billable_occurrences/{id} put on the right api version when authenticated billable_occurrence updated returns a 200 response
      Failure/Error: let(:id) { (item_class.send :create!, item_params).id }
      
      ActiveModel::UnknownAttributeError:
        unknown attribute 'attendance_params' for BillableOccurrence.
```

app/models/billable_occurrence.rb
```diff
@@ -6,6 +6,12 @@ class BillableOccurrence < UuidApplicationRecord
   belongs_to :billable, polymorphic: true
   has_many :billable_occurrence_rate_types, dependent: :destroy
   has_many :rate_types, through: :billable_occurrence_rate_types
+
+  accepts_nested_attributes_for :billable, allow_destroy: true
+
+  def build_billable(params)
```
katelovescode, line 12: This is the start of an implementation I was looking into here: https://stackoverflow.com/questions/3969025/accepts-nested-attributes-for-with-belongs-to-polymorphic/32915379 but I'm not really sure what they're suggesting you do for the controller; most of the examples I've found are for form inputs, and while they do map to attributes sometimes, it's a little tricky

db/seeds.rb
```diff
@@ -148,15 +148,17 @@ def dates_skipping_most_weekends(first_date: Date.current - 60.days,
...
 
   days_attended = dates_skipping_most_weekends(first_date: first_date, last_date: last_date)
   days_attended.each do |day_attended|
     random_checkin_time = (earliest_checkin_hour * 60) + Random.rand(60 * RAND_CHECKIN_HRS_RANGE).minutes
     random_checkout_time = random_checkin_time + Random.rand(60 * RAND_CHECKOUT_HRS_RANGE).minutes
 
-    Attendance.find_or_create_by!(check_in: day_attended + random_checkin_time,
-                                  check_out: day_attended + random_checkout_time)
+    # Attendances are a type of BillableOccurrence so we should always create them this way
+    BillableOccurrence.find_or_create_by!(child_approval: child.child_approvals[0], billable: Attendance.create!(check_in: day_attended + random_checkin_time,
```
katelovescode, line 160: This is the model of how it "feels right" to create an attendance.  It's a Billable Occurrence that belongs to a child_approval (formerly: child_case_cycle), and it's of type Attendance.

spec/requests/api/v1/billable_occurrences_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  it_behaves_like 'it creates an item', BillableOccurrence do
+    let(:item_params) { billable_occurrence_params }
+  end
+
+  it_behaves_like 'it retrieves an item for a user', BillableOccurrence do
+    let(:item_params) { billable_occurrence_params }
+  end
+
+  it_behaves_like 'it updates an item', BillableOccurrence, 'attendance_params', { "check_in": Faker::Time.between(from: Time.zone.now - 1.day, to: Time.zone.now).to_s }, nil, true do
```
katelovescode, line 24: Given the spec refactoring, I'm not sure if this is the right way to pass nested attributes but I think it's important that we test changing a nested attribute here since most of the logic lives on the subtype.

weedySeaDragon (dismissed):
LGTM!
"BillableOccurance" and related are good changes. :-)

I am not able to use swagger to "hit the api"  but testing the Pundit stuff straight in the Rails console is all working.

db/seeds.rb
```diff
@@ -167,29 +169,33 @@ def latest_date(date1, date2)
 # Attendance for Maria between January 1 and March 31
 make_attendance(first_date: JAN_1,
                 last_date: MAR_31,
-                earliest_checkin_hour: 7)
+                earliest_checkin_hour: 7,
+                child: maria)
 
 # Attendance for K'Shawn between January 1 and March 31
```
arku, line 175: **minor**: Can you update the outdated comment (K'Shawn -> Maria)?

arku (approved):

Issue #482, opened 2020-10-27
I can retrieve business from the API 


csprayregen:
## ✨ What new feature do you want to see?
The API returns the business name for the dashboard

## 🧶 Acceptance Criteria
- [ ] The `case_list_for_dashboard` endpoint returns business name
- [ ] Ticket #483 has the necessary query for this data documented

## 🖍️ Implementation Details
- Edit the /case_list_for_dashboard endpoint to return this field in the response for a list of the user's children (if it doesn't already)
- Document the endpoint and field in the ticket listed in the acceptance criteria. For example:
> Endpoint: GET /api/v1/case_list_for_dashboard
> Authentication required: yes
> Authorization: scoped to the user's children
> Example response:
> ```
> [
>     {
>         "full_name": "Maria Baca",
>         "approvals": [
>             {
>                 "case_number": "5184465331"
>             }
>         ]
>     },
>     {
>         "full_name": "K'Shawn Henderson",
>         "approvals": [
>             {
>                 "case_number": "5198908053"
>             }
>         ]
>     }
> ]
> ```

## 🥧 Additional context
Figma link for reference of initial implementation of this query on the front-end: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=2734%3A261

PR #496, opened 2020-10-29
[482] Adds business name to the dashboard endpoint

katelovescode:
Adds business name to the dashboard endpoint, adds an admin user to seeds for local manual testing

#482 

## 🍂 Before You Submit

* [X] Did you write tests?

## 🧳 Steps to test

- login via Postman
- use the token from the Authorization header in the login response in your request Authorization header as a bearer token
- hit the new endpoint: `/api/v1/case_list_for_dashboard`
- you should see the following response format, if the user you logged in with has any children associated with it (or if they are an admin user):
```
[
    {
        "full_name": "Maria Baca",
        "business": {
            "name": "Happy Seedlings Childcare"
        },
        "approvals": [
            {
                "case_number": "5302337801"
            }
        ]
    },
    {
        "full_name": "K'Shawn Henderson",
        "business": {
            "name": "Happy Seedlings Childcare"
        },
        "approvals": [
            {
                "case_number": "3124809659"
            }
        ]
    }
]
```


db/seeds.rb
```diff
@@ -26,6 +26,20 @@ def puts_records_in_db(klass)
 
 # ---------------------------------------------
 
+@user_admin = User.where(email: 'admin@test.com').first_or_create!(
+  full_name: 'Admin User',
+  greeting_name: 'Addie',
+  language: 'spanish',
+  opt_in_email: false,
+  opt_in_text: false,
+  organization: 'Pie for Providers',
```
weedySeaDragon, line 35: We're using `ENV` to get the email and password for the 1 user created, should we use `ENV` here, too?
- or only use `ENV` for the admin
- or not use `ENV` at all?

weedySeaDragon (changes requested):
Should this return only _active_ businesses or all of them?
And if all businesses, should this also return the _active_ attribute?

katelovescode:
Great question. We'll need a check on children that if their business is not active they must be turned inactive or assigned to an active business. Then this dashboard query should only return active children. I'll add that logic in.

db/seeds.rb
katelovescode: I'm gonna lean towards no ENVs, I think it makes it more confusing right now.

Issue #248, opened 2020-08-28
I can switch languages from every screen 


csprayregen:
## ✨ What new feature do you want to see?
On any screen in the app, there is a button I can click to change the language I see. 

## 🧶 Acceptance Criteria
- [x] In sizes 1024 px width or larger, I can switch languages via the button at the top right, next to my account icon. When I'm in English, this option reads "Español." When I'm in Spanish, this option reads "English." 
- [x] In sizes smaller than 1024 px, I see a hamburger menu. When I click it, I have two options - log-out, or switch your language. When I'm in English, this option reads "Español." When I'm in Spanish, this option reads "English." 

## 🖍️ Implementation Details
@katelovescode is there any additional context you want to add on the tech we're using here?

The screen sizes we're building for are below. Sizes xs and sm have the hamburger menus. Sizes md and lg have the language visible in the top right. 
-     xs: '360px',
-     sm: '768px',
-     md: '1024px',
-     lg: '1280px'

## 🥧 Additional context
Figma mockup (sizes 1024 and 1280) here **look at current state only:**: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=1165%3A0
Mobile mockup (for sizes 360 and 768) **look at current state only:**: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=1165%3A0

csprayregen:
@katelovescode @cjhaddad the hamburger menu appears in our size xs, but not in size sm. Moving back to "in progress."

cjhaddad:
@csprayregen the breakpoints in ant-design work as min-widths it seems. therefore, screens more than or equal to 768 have regular nav links and then 767 or less have hamburger. would you like it to be 768 or less? if yes, I can do some finagling. let me know! :)

csprayregen:
@cjhaddad if you can get it to be 768 or less in less than an hour that would be great! If not, we can live with it as is.

cjhaddad:
@csprayregen yes! I will look into it. may need to bypass using the built in breakpoints via ant-design, but it should be possible. @katelovescode any objections to going that route if needed?

PR #525, opened 2020-11-03
check window width for navbar and hold in state to re-render as needed

cjhaddad:
#248  💅🏼 What issue does this fix? fix nav bar responsiveness to use hamburger menu at <= 768px

## 🍂 Before You Submit

* [ ] Did you write tests?

## 🛷 Deployment

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

### DIFF
--- a/client/src/_shared/LoggedInLayout.js
+++ b/client/src/_shared/LoggedInLayout.js
@@ -70,6 +70,8 @@ export function LoggedInLayout({ children, title }) {
     )
   }
 
+  // listening for width changes of the window to make the site responsive
+  // unfortunately, ant-design breakpoints didn't include 768 <=, but 768 >=
   useEffect(() => {
     window.addEventListener('resize', setWidth)
 
--- a/client/src/_shared/LoggedInLayout.js
+++ b/client/src/_shared/LoggedInLayout.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import PropTypes from 'prop-types'
 import { useHistory } from 'react-router-dom'
 import pieSliceLogo from '_assets/pieSliceLogo.svg'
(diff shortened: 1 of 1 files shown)


client/src/_shared/LoggedInLayout.js
```diff
@@ -14,6 +14,8 @@ const { useBreakpoint } = Grid
 export function LoggedInLayout({ children, title }) {
   const dispatch = useDispatch()
   const { t, i18n } = useTranslation()
+  const [windowWidth, setWindowWidth] = useState(window.innerWidth)
```
katelovescode, line 17: 😭  Ant Design, why you make us do this

client/src/_shared/LoggedInLayout.js
```diff
@@ -24,28 +26,55 @@ export function LoggedInLayout({ children, title }) {
...
+
+    return (
+      <Dropdown overlay={mobileMenu}>
+        <MenuOutlined />
+      </Dropdown>
+    )
+  }
+
+  useEffect(() => {
+    window.addEventListener('resize', setWidth)
```
katelovescode, line 74: 😭  Here, let me reimplement browsers 😭

katelovescode (approved):
This looks great, Candice, even though it sucks we have to do it this way.  Nice work.  Do you mind adding a comment above the useEffect explaining why we had to do this for later maintenance record?

client/src/_shared/LoggedInLayout.js
cjhaddad: lol I know...it makes sense what they are doing with their breakpoints, bc they're going with min-width, but I wish that screens object was a little different. sigh

cjhaddad:
> This looks great, Candice, even though it sucks we have to do it this way. Nice work. Do you mind adding a comment above the useEffect explaining why we had to do this for later maintenance record?

Will do!

PR #542, opened 2020-11-08
issue 514: Create Child, ChildApproval from a CSV file (associates Child, SubsidyRule, Approval)

weedySeaDragon:

1. Read in an **onboarding CSV file** and from it create `Children` and `ChildApprovals` (which associates a Child with a SubsidyRule and an Approval)
2. Add a service to help with date calculations: `DateHelperService`
3. Child:  calculate the exact age in years so it can be used to find a SubsidyRule (max_age)
4. SubsidyRule:  add scopes so the right subsidy rule can be found, based on max_age and whether or not it is currently in effect
2. Added example onboarding_data.csv file, and entries in `seeds.rb` so that it can be read in

## 🍂 Before You Submit

* [x] Did you write tests?

## 🛷 Deployment

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally

1. run `rails console`
2. `> OnboardingCsvReader.import`

## 🕯 Caveats, concerns

I'm quite sure there will be many things to change.  But here is a start.

- is the name and location (service) for `DateHelperService` ok?  These methods are definitely outside of the scope for `Child` but I wasn't sure where else you might want them.


db/seeds.rb
```diff
@@ -65,44 +65,98 @@ def puts_records_in_db(klass)
...
+@biz_lil_ducks = Business.where(name: 'Lil Baby Ducklings', user: @user_kate).first_or_create(
+  license_type: Licenses.types.keys.first,
+  county: cook_county,
+  zipcode: cook_zip_60686
+)
+
 puts_records_in_db(Business)
 
+# ---------------------------------------------
+# Approvals
```
katelovescode, line 101: This seeding shouldn't be necessary, should it?  We're not referencing these approvals again, we're just creating new ones for the kids below, right?

db/seeds.rb
```diff
@@ -65,44 +65,98 @@ def puts_records_in_db(klass)
...
 # Children w/ Required Approvals
 # ---------------------------------------------
 
 # find_or_create_by! a Child with the full_name,
 #  and birthday set randomly between the min_age and max_age.
-def child_named(full_name, min_birthday: MIN_BIRTHDAY,
-                max_birthday: MAX_BIRTHDAY,
-                business: @business)
-  effective_date = Faker::Date.between(from: 1.year.ago, to: Time.zone.today)
+def child_named(full_name, date_of_birth: nil,
```
katelovescode, line 136: How come we're defaulting to nil here rather than the Faker.between?  Kids are going to be required to have dob on entry.

db/seeds.rb
```diff
@@ -116,18 +170,59 @@ def child_named(full_name, min_birthday: MIN_BIRTHDAY,
...
 # ---------------------------------------------
 
-rule_effective_date = Faker::Date.between(from: 10.years.ago, to: Time.zone.today)
+today = Date.current
 
 il_sr_rule = IllinoisSubsidyRule.first_or_create!
 
-sr_rule_1 = SubsidyRule.first_or_create!(
-  name: 'Rule 1',
+il_rules_start = today - 200
```
katelovescode, line 177: Let's name this variable something clearer since we're using it so much below. `two_hundred_days_ago`?

spec/fixtures/files/onboarding_data.csv
```diff
@@ -0,0 +1,7 @@
+first_name,last_name,date_of_birth,business_name,business_zip_code,business_county,business_qris_rating,case_number,full_days,part_days,effective_on,expires_on,co_pay,co_pay_frequency
+Juan, Oritz,2015-04-14, Happy Hearts Childcare,60606, Cook,Gold, 1234567, 18,4,2019-11-12,2020-11-12,10000,Monthly
+Julia, Oritz,2017-12-01, Happy Hearts Childcare,60606, Cook,Gold, 1234567,22,5,2019-11-12,2020-11-12,10000,Monthly
```
katelovescode, line 3: Can we create a couple records where the qris rating is nil?

app/services/onboarding_csv_parser.rb
```diff
@@ -0,0 +1,101 @@
...
+    approval_data = { case_number: row['case_number'].to_s.strip }
+    approval_data[:effective_on] = row['effective_on'] if row['effective_on'].present?
+    approval_data[:expires_on] = row['expires_on'] if row['expires_on'].present?
+    approval = Approval.find_by(approval_data)
+    raise_not_found_error("Approval case number: #{row['case_number']} effective on: #{row['effective_on']} expires on: #{row['expires_on']}") unless approval
+
+    approval
+  end
+
+  def self.create_child_approval(child, approval)
```
katelovescode, line 77: Because this info will later come in via API, I'd rater this kind of logic, creating the approval (if it doesn't already exist), associating the child_approval with the subsidy_rule and rate_types, etc., should live in a service class that can be called either by the CSV parser or the child's create and update controller methods.  I'd love to see this parser just turn the csv into a JSON object just like the API would send, and then have the CSV Reader call the service on the result to associate it with these complicated associations.

app/services/onboarding_csv_reader.rb
```diff
@@ -0,0 +1,43 @@
...
+#
+# @author Ashley Engelund (ashley.engelund@gmail.com  weedySeaDragon @ github)
+# @date   11/01/20
+#
+#--------------------------
+#
+class OnboardingCsvReader
+  IMPORT_FN = Rails.root.join('onboarding_data.csv')
+
+  def self.import(full_filename = IMPORT_FN)
```
katelovescode, line 20: I wouldn't set a default here, and I'd remove `onboarding_data.csv` from the file directory; it should accept a file as a param, and exit early/catch an error if one isn't provided.  I'll be running these data imports as individual rake tasks for unique data files every time we have to import one.

spec/mailers/previews/devise_custom_mailer_preview.rb
```diff
@@ -2,7 +2,7 @@
 
 # list these mailer preview links with this URL:
 # http://localhost:3001/rails/mailers/devise_custom_mailer
-
+#
```
katelovescode, line 5: Was this intentional?

katelovescode (changes requested):

spec/mailers/previews/devise_custom_mailer_preview.rb
weedySeaDragon: rubocop complained about the blank line there. :-)

db/seeds.rb
weedySeaDragon: They were (are) required for the `.csv` file I created.  Since I'll get rid of the csv file, I'll take them back out.

app/services/onboarding_csv_parser.rb
weedySeaDragon: Got it. 
Makes sense.

PranavChaudhary1995:
Please unsubscribe me i have tried not been able to

weedySeaDragon:
@katelovescode --  I refactored into service classes that can take JSON and create objects as needed:
- OnboardingInfoFactory  - this takes the equivalent of 1 JSON row from an onboarding CSV and finds/creates objects as necessary.  (I'm really _not_ crazy about the name.)
- ChildApprovalFactory
- SubsidyRuleFinder

I also put some of the errors into a separate file. Not sure where you would like to put it. 

Can of course name these whatever -- I am used to thinking in terms of the Factory pattern and use that.  

And I removed the index on SubsidyRule.  (We can always add it back in later, obv.)

I'm having the save issue with rubocop as in the other PR:  it doesn't like that I have a variable as the last statement in a method so that I can return it.  Perhaps it prefers the result of a conditional statement?

spec/mailers/previews/devise_custom_mailer_preview.rb
katelovescode: You could just remove the line 😂

katelovescode:
Closing this in favor of smaller, atomic PRs with service classes.

PR #551, opened 2020-11-09
issue 514:  be able to read and parse an 'onboarding' CSV => json

weedySeaDragon:

Simple file reader and parser that takes an "onboarding" CSV file and turns it into a JSON string.
This was refactored out of PR #542 

## 🍂 Before You Submit

* [x] Did you write tests?

## 🛷 Deployment

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🕯 Caveats, concerns

I'm not crazy about the names of the classes; open to suggestions!


katelovescode (dismissed):
Looks good after it passes rubocop!  Nice and tidy!

app/services/onboarding_csv_reader.rb
```diff
@@ -0,0 +1,28 @@
...
+# @author Ashley Engelund (ashley.engelund@gmail.com  weedySeaDragon @ github)
+# @date   11/01/20
+#
+#--------------------------
+#
+class OnboardingCsvReader
+  def self.import(full_filename)
+    raise ArgumentError, 'Must provide a filename' if full_filename.blank?
+
+    json = String.new  # TODO: rubocop flags this, but I don't know how else to initialize it.
```
weedySeaDragon, line 22: @katelovescode 
This is what Rubocop is complaining about.
I'm not sure how else to initialize this empty String.  (No doubt I'm having a brain fart.) Halp!

katelovescode:
```
json = ''
```
No need to initialize it as a string literal

weedySeaDragon: Nope -- that doesn't work.  That gives you a `FrozenError`:
```
 FrozenError:
       can't modify frozen String
```

weedySeaDragon: Hence my puzzlement with Rubocop.

katelovescode: Okay, this is because of the default frozen string literal setting that's recommended.  I'm near positive it's a preferred way of handling strings for performance reasons.  Could you do this instead:

```
  def self.import(full_filename)
    raise ArgumentError, 'Must provide a filename' if full_filename.blank?
    File.open(full_filename, 'r') { |csv_file| OnboardingCSVParser.parse(csv_file.read) }
  end
```

Since Ruby methods return the evaluation of the last line called?

weedySeaDragon: Yup - that's why.  (It's a 'performance' cop.)
I've already refactored to a slightly different way that is passing.
Let me know what you think & if you want me to use your suggestions.

.rubocop.yml
```diff
@@ -39,6 +39,9 @@ Lint/StructNewOverride:
 
 Rails:
   Enabled: true
+  Exclude:
+    - app/services/onboarding_csv_reader.rb
```
katelovescode, line 43: Rude 😂 all kidding aside, I don't want to exclude these because I expect them to conform to the same behavior as the rest of the app, what issues were you getting?

spec/services/onboarding_csv_reader_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require_relative File.join(__dir__, '..', 'spec_helper')
```
katelovescode, line 3: This shouldn't be necessary; having the services in the app file means they'll be autoloaded, and RSpec should be looking in the directory because of the names.  If that's not working I can do a little digging on this.

katelovescode (commented):
Just a couple of comments left.

app/services/onboarding_csv_parser.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require 'csv'
+require_relative 'onboarding_csv_parser'
```
katelovescode, line 4: Same here - this class should be autoloaded without require_relative

app/services/onboarding_csv_reader.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'csv'
+require_relative 'onboarding_csv_parser'
```
katelovescode, line 4: ditto

spec/services/onboarding_csv_parser_spec.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require_relative File.join(__dir__, '..', '..', 'app', 'services', 'onboarding_csv_parser')
```
katelovescode, line 4: ditto

katelovescode (approved):

katelovescode (approved):

Issue #217, opened 2020-08-23
There is a secondary scroll bar in the log in and create an account pages 


csprayregen:
## 🤬 What's wrong?
There is a secondary scroll bar in the log in and create an account pages 
Relatedly, the image should be fixed, while the right side should scroll, but they are both scrolling currently. 

## 👉🏼 Steps to Reproduce

- go to `/login` at full desktop size (lg. or larger)

## 🧥Expected behavior
We want the image to be stationary and the right side to scroll.
Only one scroll bar should display, for the right side, rather than two stacked scrollbars

csprayregen:
Hi @katelovescode I know this came up at our meeting last Sunday, but I'm not able to reproduce this error. 
On Chrome, I do see that the image is scrolling rather than fixed, but I don't see the secondary scroll. This feels like a low priority fix to me. 
On Safari, it's all working as expected. 
What are you experiencing?

shrmpy:
I'm interested in this. Just managed to build and start the web frontend, and can see double scrollbars on the Sign Up page with the left column image scrolling along as the right column form. I'm viewing from Chromium on Linux.

shrmpy:
I have the change to client/src/_shared/AuthLayout.js and would like to get feedback. Specifically, the AuthLayout is shared by Signup,Login,Password/update so this change potentially affects the UI of these other pages. 

I don't have a UI test yet. I did make sure that I did not break the existing tests [frontend](https://app.circleci.com/pipelines/github/shrmpy/pieforproviders/70/workflows/357c7d80-f3d1-4450-b0ea-505cf99e0ca7/jobs/135/parallel-runs/0/steps/0-107) [e2e](https://app.circleci.com/pipelines/github/shrmpy/pieforproviders/70/workflows/357c7d80-f3d1-4450-b0ea-505cf99e0ca7/jobs/136/parallel-runs/0/steps/0-121).

Proposed change attached here (client/src/_shared/AuthLayout.js):
[AuthLayout.txt](https://github.com/pieforproviders/pieforproviders/files/5420368/AuthLayout.txt)

katelovescode:
Shrmpy - we won't review changes that don't come in a pull request - I'm going to unassign this so others can take it, but feel free to submit a pull request if you can.

shrmpy:
[...]

PR #590, opened 2020-11-21
217/secondary scroll bar bugfix

cjhaddad:
#217 fixes multiple scrollbars that show up on `/login` and `/setup` screens that shows up on some browsers
also fixes scrolling issue found on same routes on chrome

## 🍂 Before You Submit

* [ ] Did you write tests?

## 🛷 Deployment

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

### DIFF
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -3,7 +3,8 @@ import {
   BrowserRouter as Router,
   Redirect,
   Route,
-  Switch
+  Switch,
+  useLocation
 } from 'react-router-dom'
 import AuthenticatedRoute from '_utils/_routes/AuthenticatedRoute.js'
 import ReactGA from 'react-ga'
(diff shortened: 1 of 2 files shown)


client/src/App.js
```diff
@@ -20,10 +21,62 @@ import { AuthLayout } from '_shared'
...
-const App = () => {
+const Routes = () => {
   const { t } = useTranslation()
   const isAuthenticated = useAuthentication()
+  let location = useLocation()
+
+  return (
+    <div
+      className={`text-primaryBlue font-proxima text-sm h-screen ${
+        location.pathname === '/signup' || location.pathname === '/login'
```
katelovescode, line 32: wow wow wow 😭 this picture is going to be the death of us; rather than grabbing by location pathname, is there any chance we could do "if the currently-rendered component is wrapped in an AuthLayout, overflow-hidden, else don't do anything"?  Or is that too meta?

client/src/App.js
```diff
@@ -20,10 +21,62 @@ import { AuthLayout } from '_shared'
 import { useTranslation } from 'react-i18next'
 import { useAuthentication } from '_shared/_hooks/useAuthentication'
 
-const App = () => {
+const Routes = () => {
```
katelovescode, line 24: Nice cleanup here <3

katelovescode (approved):
Just a tiny comment but if the thing I'm proposing is overkill, then ignore it, I'm good with this going in as-is!

client/src/App.js
cjhaddad: yes, I remember hearing rumblings re: this picture. Similarly, I had the idea of inserting this `overflow-hidden` class into the top-level div in the AuthLayout component, but unfortunately it wasn't top level enough to have the effect we need. let me see/ think if there is a way to identify AuthContent as the component rendered for the conditionally here, bc this top-level div is the div we need to add the class.

Issue #561, opened 2020-11-16
Convert the SubsidyRuleAssociator to a background job


katelovescode:
## 📋 Summary
When we import CSVs we'll be associating a lot of children to SubsidyRules, we should do that as a background job

## 💻 Implementation Details

- I'd like to investigate Resque before I settle on Sidekiq

## 🧐 Why tho

Speed

PR #631, opened 2020-12-05
Move subsidy rule association to a background job

arku:

https://github.com/pieforproviders/pieforproviders/issues/561

## 🍂 Before You Submit

* [ ] Did you write tests?

## 🛷 Deployment

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test

Run `foreman start` and create a child with a business in IL from from the rails console.


katelovescode (changes requested):
This is a good start, couple implementation tweaks:
- can we enable the dashboard engine? https://github.com/bensheldon/good_job#execute-jobs-async--in-process
- can we run jobs async with the web server rather than a separate process?  I know Heroku resources get expensive fast: https://github.com/bensheldon/good_job#execute-jobs-async--in-process

arku:
> can we enable the dashboard engine? 

I'm not sure how to do it with our current setup. We only want to allow admin users to access `/good_job`. GoodJob's documentation suggests using Devise's `authenticate` helper to authorize access but we can't use it since we don't use session cookies.

```
authenticate :user, ->(user) { user.admin? } do
  mount GoodJob::Engine => 'good_job'
end
```

How do we go about it?

katelovescode:
> How do we go about it?

Could we write a non-api controller, inherit from GoodJob's dashboards controller, and apply a Pundit policy to it? https://github.com/bensheldon/good_job/blob/main/engine/app/controllers/good_job/dashboards_controller.rb

arku:
> > How do we go about it?
> 
> Could we write a non-api controller, inherit from GoodJob's dashboards controller, and apply a Pundit policy to it? https://github.com/bensheldon/good_job/blob/main/engine/app/controllers/good_job/dashboards_controller.rb

We could do that. This would require us to create another controller + view for login. A simpler solution would be to use basic auth. WDYT?

katelovescode:
> We could do that. This would require us to create another controller + view for login. A simpler solution would be to use basic auth. WDYT?

Personally I'd rather maintain another controller (I think the view will be built in with the good_job gem, we shouldn't have to create one to inherit, as long as the methods use `super`, we should just be able to scope it?  I think?) than a second authentication method.

arku:
> Personally I'd rather maintain another controller (I think the view will be built in with the good_job gem, we shouldn't have to create one to inherit), and in the controller, as long as the methods use super, we should just be able to scope it? I think?

@katelovescode Yeah, that's right about the good job controller. The second controller I was referring to is the `LoginController` to render the login form view. Note that we're already overriding the devise sessions controller and rendering JSON.

arku:
> All of that sounds better than maintaining a second auth structure

I was suggesting basic auth to authorize access as mentioned in https://github.com/bensheldon/good_job#dashboard to keep the code simple for now. If we do want to add new views for login, we can always do it in the future.

katelovescode:
> > Personally I'd rather maintain another controller (I think the view will be built in with the good_job gem, we shouldn't have to create one to inherit), and in the controller, as long as the methods use super, we should just be able to scope it? I think?
> 
> @katelovescode Yeah, that's right about the good job controller. The second controller I was referring to is the `LoginController` to render the login form view. Note that we're already overriding the devise sessions controller and rendering JSON.

Can we co-opt our existing login, though?  Like, can we route to the GoodJob views via React and using the auth token?

katelovescode (dismissed):

katelovescode (approved):

Issue #140, opened 2020-07-04
Implement typography/font styles


rebeccakarasiktw:
## 📋 Summary
There was a mis-translation between designs in Figma and development. We want to make sure all H1, H2, etc is consistent across the app, and takes into account responsive sizes as well. 

## 💻 Implementation Details
Implement the styles according to Figma: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=175%3A1984
Update headers across the pages that have already been published to ensure they are using the correct font styles.

## 🧐 Why tho
Want to make sure we create a consistent user experience of course! And also to make us faster using consistent styles.

katelovescode:
@jontrainor this ticket should go on hold, it's blocked behind some new design work.  Rebecca is going to rethink the font sizes in `rem` instead of explicit pixel sizes.

Jess-White:
Hi - This seems like a great way to familiarize myself with the codebase - can I jump on it? I know it's hacktoberfest-tagged; I can look for other options. Thanks!

jontrainor:
Yes! I was waiting for Rebecca to have some time to chat about it because
there was "large" and "small" typography sets and Ant just wants all font
sizes relative to the base font size. I would touch base with her but go
for it! I've had a busy month and haven't followed up on it.

Jess-White:
Okay, I'm on the case!

rebeccakarasiktw:
Hello!! I'm very sorry I totally missed this message. @Jess-White let me know how I can help!

rebeccakarasiktw:
Hi @cjhaddad I noticed a few small fonts that should be updated, I've added them as screenshots here, lmk if you have any questions!
![Group 255](https://user-images.githubusercontent.com/26929739/108082096-b3662b00-703f-11eb-98b6-d972e0d9c327.png)
![Frame 81](https://user-images.githubusercontent.com/26929739/108082102-b4975800-703f-11eb-8ee4-64ea4f85151e.png)
![Frame 80](https://user-images.githubusercontent.com/26929739/108082104-b52fee80-703f-11eb-897f-d087d50a420c.png)

rebeccakarasiktw:
Hi @cjhaddad I took a look and it all looks good! Thanks!

PR #815, opened 2021-01-15
Task/140/implement typography font styles

Jess-White:

#140 
- customize tailwind.config.js file to reflect values in figma typography spreadsheet
- create classes in tailwind.css file to include those values, also using figma typography spreadsheet
- add classes inline to components using mockups for each component in figma (<----partly completed)

## 🍂 Before You Submit

* [ ] Did you write tests?

## 🛷 Deployment

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🕯 Caveats, concerns

- did not finish adding inline styles to components


katelovescode (changes requested):
Looks like we accidentally committed a `.DS_Store` file and some older files that are unrelated to this PR, can we rebase and remove the `.DS_Store`?  Probably need to do a `bfg` clean so it doesn't get recommitted: https://rtyley.github.io/bfg-repo-cleaner/

We should remove db/structure.sql from this commit, and the yarn.lock might be different after a rebase.

The only other note is that it doesn't look like there are changes here for the actual font-face, and I think I implemented the font-face wrong in the first place.  On the typography ramp in Figma, there are different font-faces, not just different weights, can we do an audit of that?  If nothing needs to be changed, just a comment and a 👍🏼  from Rebecca would be great - I'm like 99% positive I messed it up when I did it in the summer, so I just want to make sure a person who is not me and understands type looks at it :D

katelovescode (approved):
Tested locally, inspected the H3 on the Login page and it's Proxima Nova Alt 🎉

Issue #677, opened 2020-12-12
As a NE user, I can see dynamic data in the "as of" date field


csprayregen:
## ✨ What new feature do you want to see?
In the NE dashboard, I see a current "as of" date. This is defined as the last date attendance was uploaded for any child.

## 🧶 Acceptance Criteria
- [x]  I see the As Of date on my dashboard 
- [x]  The above criterion works in both English and Spanish
- [x]  Verify accessibility (see details below)

## 🖍️ Implementation Details
[
    {
        "absences": "1 of 5",
        "approvals": [
            {
                "case_number": "4441500188"
            }
        ],
        "as_of": "12/21/2020",
        "business": {
            "name": "Nebraska Home Child Care"
        },
        "case_number": "case#1",
        "earned_revenue": 1045.32,
        "estimated_revenue": 2022.14,
        "full_days": "14 of 15",
        "full_name": "Rhonan Shaw",
        "hours": "1 of 6.5",
        "transportation_revenue": "30 trips - $80.00"
    },
    {
        "absences": "1 of 5",
        "approvals": [
            {
                "case_number": "8712853078"
            }
        ],
        "as_of": "12/21/2020",
        "business": {
            "name": "Nebraska Home Child Care"
        },
        "case_number": "case#1",
        "earned_revenue": 1045.32,
        "estimated_revenue": 2022.14,
        "full_days": "14 of 15",
        "full_name": "Tanim Zaidi",
        "hours": "1 of 6.5",
        "transportation_revenue": "30 trips - $80.00"
    }
]

> **NOTE TO ENGINEERS**
> 
> Please ensure that you are running Google Lighthouse and/or WebAIM (Wave) on any UI elements in your pull request to check for accessibility issues.
> 
> If it is out of scope to address those accessibility issues in your pull request (because the issue is in a parent component and needs refactoring to work, for example), **you are responsible for noting those issues in your pull request and writing bug tickets to address the issues - make sure you link the new tickets in your pull request**

## 🥧 Additional context
Mockup: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=3579%3A1
Spanish: https://docs.google.com/spreadsheets/d/1d0RPc9PXvTCFpq8bfkZxgNL35MEpX5s8EvBwqex8564/edit#gid=0

cjhaddad:
in the case list for the dashboard this will be a new key "as_of": dateString

csprayregen:
[...]

PR #831, opened 2021-01-22
configure as of figure and static date filter number and render

cjhaddad:
#677

## 🍂 Before You Submit

* [ ] Did you write tests?

## 🛷 Deployment

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test
login with both IL and NE users and confirm that a date filter button with static data showing the current month renders in the dashboard title and an 'as of' data point configured from the user data from API

### DIFF
--- a/client/src/Dashboard/Dashboard.js
+++ b/client/src/Dashboard/Dashboard.js
@@ -278,7 +278,7 @@ export function Dashboard() {
           <Typography.Title className="dashboard-title mr-4">
             {t('dashboardTitle')}
           </Typography.Title>
-          <Button className="date-filter-button mr-2 text-base py-2 px-4">
+          <Button className="date-filter-button mr-2 text-base py-2 px-4" disabled>
             {dates.dateFilter}
           </Button>
           <Typography.Text className="text-gray3">{`As of: ${dates.asOf}`}</Typography.Text>
--- a/client/src/Dashboard/Dashboard.js
+++ b/client/src/Dashboard/Dashboard.js
@@ -200,34 +200,8 @@ export function Dashboard() {
   }
 
   const reduceAsOfDate = res => {
-    const practice = [
-      {
-        as_of: '01/21/2020',
-        businesses: Array(1),
-        max_revenue: 23122,
-        total_approved: 23122
-      },
-      {
-        as_of: '01/21/2021',
-        businesses: Array(1),
-        max_revenue: 23122,
-        total_approved: 23122
-      },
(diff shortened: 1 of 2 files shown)


client/src/Dashboard/Dashboard.js
```diff
@@ -252,7 +274,15 @@ export function Dashboard() {
   return (
     <div className="dashboard sm:mx-8">
       <div className="dashboard-title m-2">
-        <Typography.Title>{t('dashboardTitle')}</Typography.Title>
+        <div className="flex items-center mb-3">
+          <Typography.Title className="dashboard-title mr-4">
+            {t('dashboardTitle')}
+          </Typography.Title>
+          <Button className="date-filter-button mr-2 text-base py-2 px-4">
```
katelovescode, line 281:
```suggestion
          <Button className="date-filter-button mr-2 text-base py-2 px-4" disabled>
```
@rebeccakarasiktw might want this button disabled until it actually does something, in the spirit of iterative development.  That means the styling will be gray though, so that might not be desired.  If she's okay with it, I'm 👍🏼

katelovescode (changes requested):
Seeing a regression on the sign up/login buttons
Screen Shot 2021-01-24 at 2 12 11 PM

client/src/Dashboard/Dashboard.js
cjhaddad: makes sense to me to just go ahead and add the disabled attribute until it does something! will add.

cjhaddad:
> Seeing a regression on the sign up/login buttons
> Screen Shot 2021-01-24 at 2 12 11 PM

lol wtf, looking into this!

client/src/Dashboard/Dashboard.js
cjhaddad: just checked this out -- adding the disabled attr adds a bunch of ant-design styling. the way I have it right now, the button is essentially disabled, bc there is no hover, active effect and no associated onClick event. it just looks like a static rectangle UI. If we want to add the 'disabled' attr, I am sure I can figure out how to override the styles.

cjhaddad:
@katelovescode I accidentally deleted your comment! 😦 lol, anyway, I added disabled attr to the button and overrode the ant-design styles for disabled buttons, so now it just looks like a simple rectangle, per the mockup (cc: @rebeccakarasik

katelovescode (approved):

Issue #674, opened 2020-12-12
I can see definitions on my NE dashboard 


csprayregen:
## ✨ What new feature do you want to see?
Most columns have a question mark hover component. When I click on it, the definition section expands at the bottom. 

## 🧶 Acceptance Criteria
- [x] There is a functional question mark in the full days column
- [x] When I click the full days question mark, the definition section opens at the bottom and it matches the final definition in the mockup
- [x] There is a functional question mark in the hours column
- [x] When I click the hours question mark, the definition section opens at the bottom and it matches the final definition in the mockup
- [x] There is a functional question mark in the absences column
- [x] When I click the absences question mark, the definition section opens at the bottom and it matches the final definition in the mockup
- [x] There is a functional question mark in the earned revenue column
- [x] When I click the earned revenue question mark, the definition section opens at the bottom and it matches the final definition in the mockup
- [x] There is a functional question mark in the estimated revenue column
- [x] When I click the estimated revenue question mark, the definition section opens at the bottom and it matches the final definition in the mockup
- [x] There is a functional question mark in the transportation revenue column
- [x] When I click the transportation revenue question mark, the definition section opens at the bottom and it matches the final definition in the mockup
- [x] The "Back to Top" button takes me back to the top 
- [x] The above criteria work in English and Spanish
- [x] Check for accessibility

## 🖍️ Implementation Details
@rebeccakarasiktw can you add details about the Ant components here and link to the final definitions when they are ready? 
**Ant Design Details**

1. For the definitions, the component is "Collapse - Borderless" https://ant.design/components/collapse/#header
There is only 1 component
Upon clicking the "functional question mark" icon, the user should jump to that definition, with the collapse component open by default.

2. For the back to top button, the component is "Button": https://ant.design/components/button/
It should include the "Icon" property, and the size property of "large"
Also, the icon to be used is the arrow pointing up: [Here](https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=423%3A11) it is in the Style Guide, and you should be able to export it as an SVG. 

> **NOTE TO ENGINEERS**
> 
> Please ensure that you are running Google Lighthouse and/or WebAIM (Wave) on any UI elements in your pull request to check for accessibility issues.
> 
[...]

PR #887, opened 2021-02-05
674/ne definitions

cjhaddad:
#674 definitions panel for NE dashboard

## 🍂 Before You Submit

* [X] Did you write tests?

## 🛷 Deployment

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test
the #674 has a checklist that should cover all the aspects of this ticket.


client/src/i18n/locales/en/index.json
```diff
@@ -185,5 +185,17 @@
...
+  "dec": "Dec",
+
+  "definitions": "Definitions",
+  "exceededLimitDef": "The child has attended more days than they were approved for this month. You have earned the max monthly payment and will not be paid for additional days.",
+  "onTrackDef": "The child is attending most or all of their days and you will likely receive most of the subsidy payment this month.",
+  "atRiskDef": "The child has been absent frequently and you may not receive the full subsidy payment this month.",
+  "fullDaysDef": "Days with 6-10 hours attended. Approved days come from the approval letter.",
+  "hoursDef": "Days with less than 6 hours attended.",
+  "absencesDef": "You can be paid for up to five absences. Absences are included in your revenue calculations.",
+  "earnedRevenueDef2": "The revenue you have already earned this month, based on the child’s attendance. This does not include other subsidy revenue such as transportation.",
```
katelovescode, line 197: Can we give this block of definitions (and probably the ones from IL) more meaningful names? `neEarnedRevenueDef` and `ilEarnedRevenueDef` or something similar to that?  Would it make sense (at some point, not necessarily in the scope of this ticket) to refactor these locale files to have a more nested structure, i.e. group universal text blocks under "universal", then nest all the nebraska definitions under a "nebraskaDefinitions" key, etc.?

katelovescode (changes requested):
Function name typo, suggestion for translation key names.  Behavior-wise, do we want the clickable question mark to also sort the column at the same time? It might be hard to make it........not behave that way, but it's maybe a UX thing to consider, cc @rebeccakarasiktw would love to get your input.

katelovescode:
@rebeccakarasiktw this is the behavior on this branch

sorting

client/src/i18n/locales/en/index.json
cjhaddad: yes! I like the idea of the refactor. perhaps, we can create a ticket for that. in the meantime, I will remain these to be 'ne' specific!

cjhaddad:
> @rebeccakarasiktw this is the behavior on this branch
> 
> sorting

Yes! I know this is happening and did not want/ like this. I futzed with it for awhile and planned to bring it up at our Sunday meeting. The sorting UI is all happening via the ant-design components. The column header title we have control over, but it is wrapped in the ant-design elements for the sorting, which is ANNOYING, bc it wraps it in the a element that I can't figure out how to get access to. I don't know if this is making sense...but I will say more in our touchbase.

katelovescode (approved):
Amazing!  Looks great!

Issue #928, opened 2021-02-14
I can select a different month to view on my dashboard 


csprayregen:
## ✨ What new feature do you want to see?
I can click the dropdown at the top of my dashboard to toggle between months. I can view any month for which my account has data. 

## 🧶 Acceptance Criteria
- [x] The month "box" next to the words "Your Dashboard" opens up to a dropdown 
- [x] I can click on any previous month and see my dashboard as of the last day of that month 
- [x] The earliest month on the dropdown is the earliest start date of an approval for a kid on my account 
- [x] This works in English and Spanish
- [ ] This works for Nebraska users --> currently hidden until we move to the fully automated MVP
- [x] This works for IL users 

## 🖍️ Implementation Details
The API call is ticketed in #846 

> **NOTE TO ENGINEERS**
> 
> Please ensure that you are running Google Lighthouse and/or WebAIM (Wave) on any UI elements in your pull request to check for accessibility issues.
> 
> If it is out of scope to address those accessibility issues in your pull request (because the issue is in a parent component and needs refactoring to work, for example), **you are responsible for noting those issues in your pull request and writing bug tickets to address the issues - make sure you link the new tickets in your pull request**

## 🥧 Additional context
Figma mockup: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=4547%3A16651
Spanish: https://docs.google.com/spreadsheets/d/1d0RPc9PXvTCFpq8bfkZxgNL35MEpX5s8EvBwqex8564/edit#gid=0

csprayregen:
@cjhaddad @katelovescode check out the second AC above - is this right?

csprayregen:
Default cases:
1. As of date will be the last date that attendance was uploaded for the month that I"m looking at. 
2. For each month, the only kids listed on my grid are the ones that had active approvals in that month

Potential edge cases:
1. If I toggle back to a month before a kid's case was active, does that kid appear on the dashboard at all? (Will likely want to add logic that takes them away as inactive in a month that does not fall into their approval). We think the kid will be in there, and there will be zeroes across the board. 
2. If I look at a month where there are zero attendances for a kid, but their approval is active, the as of date will appear as the current date, rather than a date in the month that I'm looking at. Change to say last date of the month that I'm looking at. 
-KD: as of date is whatever date attendance

Issue #846, opened 2021-01-27
When I select a month in the date dropdown, that month is requested from the API


rebeccakarasiktw:
## ✨ What new feature do you want to see?
When I select a month in the date dropdown, that month is requested from the API. The user interaction with the dropdown is ticketed in #928 

## 🧶 Acceptance Criteria
- [x] When the date is changed, and the new data is retrieved, the front end re-renders with new data 
- [x] This works for the NE and IL dashboard --> IL only for now; NE is not visible

## 🖍️ Implementation Details

![Screen Shot 2021-03-07 at 3 49 01 PM](https://user-images.githubusercontent.com/8364647/110256107-baa49880-7f5c-11eb-9433-abfc712b2c30.png)

## 🥧 Additional context
Figma mockup: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=4547%3A16651
Spanish: https://docs.google.com/spreadsheets/d/1d0RPc9PXvTCFpq8bfkZxgNL35MEpX5s8EvBwqex8564/edit#gid=0

cjhaddad:
@katelovescode a reminder to add what the name of the data field is

katelovescode:
@cjhaddad - the value of the `first_approval_effective_date` field can be nil or a date

cURL request:
```
curl --location --request GET 'localhost:3000/api/v1/case_list_for_dashboard?filter_date=2021-01-02' \
--header 'Accept: application/vnd.pieforproviders.v1+json' \
--header 'Authorization: Bearer <token>'
```

PR #1055, opened 2021-03-16
846/dropdown date filter

cjhaddad:
This PR addresses for adding datefilter functionality:
#846 and #928 

## 🍂 Before You Submit

* [X] Did you write tests?

## 🛷 Deployment

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧵 Steps to set up locally
This PR adds a date filtering dropdown to the dashboard view for IL users. NE users should still see only a disabled button with the current month in it. 

The dropdown uses the 'first_effective_approval_date` to determine the first month of the dropdown. All subsequent months from the first month to the present are shown as options in the dropdown. Once a new month is selected the frontend should make a new call to the api's dashboard endpoint and re-render the the dashboard with the updated data for the selected month. The month currently selected should now be the month the user selects and has data showing in the dashboard table and stats.


katelovescode:
Some of this might be an issue with backend responses or front end filtering, I'm not sure yet, but a couple weird things:

- when I land on the page the first time, the dropdown displays the *oldest* date rather than today - in this case Aug 2019:

    Screen Shot 2021-03-17 at 8 40 34 AM

- it also shows unexpected data - for example, these are ALL the kids under this user (from seeding), including ones that don't have an approval for August of 2019 - Tarquinius Kelley's approvals don't go into effect until September 2019:

    Screen Shot 2021-03-17 at 8 37 30 AM

- when I click the "Aug 2019" again, it doesn't do anything, but when I click away to another month, and then back to Aug 2019, it seems to load the correct data for August - just Sabina is on the list now:

    dropdown

- clicking the arrow doesn't consistently work - from the best i can tell, it only works when I've navigated to another browser tab and come back:

    > not working: 
    > arrowdoesntwork
    > 
    > working after I navigate away:
    > arrowworksafterswitchingtabs

Other than this behavior though, the data polling by date does seem to be working!  Let me know if I need to make fixes on the backend for this if you get a chance to investigate; I can also take a look a little later this weekend.

katelovescode (changes requested):
See my notes about actual behavior; haven't reviewed the code yet.

client/src/Dashboard/DashboardTitle.js
```diff
@@ -0,0 +1,81 @@
...
+
+  return (
+    <div className="dashboard-title m-2">
+      <div className="flex items-center mb-3">
+        <Typography.Title className="dashboard-title mr-4">
+          {t('dashboardTitle')}
+        </Typography.Title>
+        {userState !== 'NE' ? (
+          <Select
+            // suffixIcon={
```
katelovescode, line 36: Booooooo antd

katelovescode (approved):

PR #1160, opened 2021-04-11
fix: Replace compile-time env vars with runtime env vars

arku:

[When the Staging app is promoted to Production, the app isn't rebuilt](https://devcenter.heroku.com/articles/pipelines#deployment-with-pipelines) and so the frontend JS bundle with the environment variables from Staging gets deployed to Production.  This PR fixes the issue by replacing compile-time variables with runtime ones. This also has the added benefit of updating the app immediately when the environment variables get changed. To get this to work, we should do the following steps:

1. Add `create-react-app-inner-buildpack` to the Production app buildpacks. This buildpack enables runtime environment variables.

```bash
heroku buildpacks:add -i 1 https://github.com/mars/create-react-app-inner-buildpack
```

2. Add the `JS_RUNTIME_TARGET_BUNDLE` environment variable to point to the JS bundle location for runtime config injection.

```
heroku config:set JS_RUNTIME_TARGET_BUNDLE='public/static/js/*.js'
```

Check [this](https://github.com/mars/heroku-js-runtime-env#how-does-it-work) out if you want to understand how it works.

## 🍂 Before You Submit

* [ ] Did you write tests?

## 🛷 Deployment

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test

1. Deploy the app to Staging and promote it to Production.
2. Verify that the build uses environment variables from Production.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -10,6 +10,7 @@
   "dependencies": {
     "@appsignal/javascript": "^1.3.12",
     "@craco/craco": "^5.9.0",
+    "@mars/heroku-js-runtime-env": "^3.0.2",
     "@material-ui/core": "^4.11.3",
     "@material-ui/icons": "^4.11.2",
     "@reduxjs/toolkit": "^1.5.1",
--- a/client/src/App.js
+++ b/client/src/App.js
@@ -7,6 +7,7 @@ import {
   useLocation
 } from 'react-router-dom'
 import { useSelector } from 'react-redux'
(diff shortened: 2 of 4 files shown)


cjhaddad (approved):

cjhaddad:
Thank you @arku this looks good to me! For the two steps you mention above -- those both need to be run for the pie-prod environment only?

arku:
@cjhaddad We should run those on both Staging and Production apps.

Issue #777, opened 2021-01-07
Tweak copy on IL dashboard


csprayregen:
## 🧶 Acceptance Criteria
- [x] Change "not met" tag to "won't meet"
- [x] Update Spanish translation "won't meet" = "no cumplirá"
- [x] Change "Guaranteed revenue" on rollup to "Earned revenue"
- [x] Change "Guaranteed revenue" header on dashboard table to "Earned revenue"
- [x] Update Spanish translation on rollup "Earned revenue" = "Ingresos ganados"
- [x] Update Spanish translation on table "Earned revenue" = "Ingresos ganados"
- [x] Check for accessibility (see below)

## 🖍️ Implementation Details

> **NOTE TO ENGINEERS**
> 
> Please ensure that you are running Google Lighthouse and/or WebAIM (Wave) on any UI elements in your pull request to check for accessibility issues.
> 
> If it is out of scope to address those accessibility issues in your pull request (because the issue is in a parent component and needs refactoring to work, for example), **you are responsible for noting those issues in your pull request and writing bug tickets to address the issues - make sure you link the new tickets in your pull request**

## 🥧 Additional context
Mockup: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=4109%3A16322
Spanish: https://docs.google.com/spreadsheets/d/1d0RPc9PXvTCFpq8bfkZxgNL35MEpX5s8EvBwqex8564/edit#gid=0

PR #1199, opened 2021-04-27
Tweak dashboard language

wittejm:
* rename `notMet` variables to `willNotMeet` to match language change. But I wanted to avoid using "won't" in a variable name. 
 * leave `guaranteedRevenue` variable names as-is because the language is still used in the data structure for Illinois. The language files now have two strings `earnedRevenue` and `guaranteedRevenue` with the same value for both English and Spanish. This is possibly subject to a bigger refactor. 
 * I don't know how to write example data that will show the "won't meet" label, so I couldn't visually check that the label looks nice. It's a small tweak though, so maybe I can figure out how to write new test data later :+1: 

#777 
## 🍂 Before You Submit

* [x] Did you write tests? (existing test updated)
* [x] Did you run Google Lighthouse and/or WebAIM (Wave) on UI components in your PR? -- WebAIM is giving some unrelated alerts on the Dashboard unaffected by this PR. Tickets forthcoming if they don't exist yet...
* [x] Does your PR contain any required translations? 
* [x] Are your primary keys UUIDs on any new tables? (n/a)

## 🕯 Caveats, concerns

(in description above)


katelovescode:
This looks good to me!  I can get you some steps on how to update the dashboard but I think I need to make it a little less coupled to Amazon S3 right now so that local development is less of a pain.  I'll be doing some architecture work tomorrow so hopefully I'll have a better idea of how to manage dashboard test cases soon.

katelovescode (approved):

Issue #1189, opened 2021-04-22
Backend Work for Mark Inactive (Historical Data)


cjhaddad:
## 📋 Summary
For frontend tickets: #1124 #1174 we need to make sure the child update endpoint takes in the following fields and adds them to the database:
```
{
   active: boolean,
   inactive_date: YYYY-MM-DD,
   inactive_reason: enum [ no_longer_in_my_care, no_longer_recieving_subsidies, other]
}
```

Also! We need to update the response data from the cases_list_for_backend to add the `active` field for each cases object from the child model.

This ticket along with #1206 should take care of the backend work needed for the mark inactive feature.

## 💻 Implementation Details

**Not in scope, but related**
* how are we going to manage historical data?
* how will we retrieve historical data?
* do we need to store data after the case becomes inactive?

cc @csprayregen 

## 🧐 Why tho
We need this change in order to make a child as inactive and active.

csprayregen:
FYI @cjhaddad and @katelovescode I've added these open questions to my to do list to scope for the backlog

katelovescode:
@cjhaddad this will need your QA/Approval as well! ✅

PR #1221, opened 2021-05-03
New Child fields for active status

wittejm:

#1189
## 🍂 Before You Submit

* [x] Did you write tests? 
Just one: for new scope `not_deleted` on child. All other changes handled by updates to other tests.
* [x] Did you run Google Lighthouse and/or WebAIM (Wave) on UI components in your PR? **N/A**
* [x] Does your PR contain any required translations?  **no**
* [x] Did you run rake prep?
* [x] Are your primary keys UUIDs on any new tables?  **N/A**

## 🛷 Deployment

* [ ] Data Migrations
* [x] Schema Migrations
* [ ] Dependencies

## 🧳 Steps to test

Test child PUT endpoint in postman
## 🕯 Caveats, concerns

new date field is `last_active_date` instead of the `inactive_date` listed in the issue #1189


app/models/child.rb
```diff
@@ -20,10 +20,19 @@ class Child < UuidApplicationRecord
   # this will try to validate with the incorrect dob even though the record has already failed
   validates :full_name, uniqueness: { scope: %i[date_of_birth business_id] }, unless: -> { errors }
 
+  REASONS = %w[
+    no_longer_in_my_care
+    no_longer_recieving_subsidies
+    other
+  ].freeze
+  validates :inactive_reason, inclusion: { in: REASONS }, allow_nil: true
```
katelovescode, line 28:
```suggestion

  validates :inactive_reason, inclusion: { in: REASONS }, allow_nil: true
```

katelovescode: Just to add a space for readability

spec/models/child_spec.rb
```diff
@@ -56,6 +56,7 @@
 
   context 'scopes' do
     let(:inactive_child) { create(:child, active: false) }
+    let(:deleted_child) { create(:child, deleted: true) }
 
     it 'only displays active children in the active scope' do
       expect(Child.active).to include(child)
```
katelovescode, line 62:
```suggestion
      expect(Child.active).to include(child)
      expect(Child.active).to not_include(deleted_child)
      expect(Child.active).to not_include(inactive_child)
```

spec/models/child_spec.rb
```diff
@@ -66,6 +67,11 @@
       expect(Child.approved_for_date(child.approvals.first.effective_on)).to include(child)
       expect(Child.approved_for_date(child.approvals.first.effective_on - 1.day)).to eq([])
```
katelovescode, line 68:
```suggestion
      expect(Child.approved_for_date(child.approvals.first.effective_on)).to not_include(inactive_child)
      expect(Child.approved_for_date(child.approvals.first.effective_on)).to not_include(deleted_child)
```

spec/models/child_spec.rb
```diff
@@ -66,6 +67,11 @@
...
     end
+
+    it 'displays inactive children but not deleted children in the not_deleted scope' do
+      expect(Child.not_deleted).to include(inactive_child)
```
katelovescode, line 72:
```suggestion
      expect(Child.not_deleted).to include(child)
      expect(Child.not_deleted).to include(inactive_child)
```

spec/models/child_spec.rb
```diff
@@ -56,6 +56,7 @@
```
katelovescode, line 56: Along with the validation specs above (line 13 - 48) can we add a validation spec for the inclusion validator, for example:
```
expect(subject).to validate_inclusion_of(:active).in_array(%w[Y N])
```

Also could you duplicate the date_param `it` block for `last_active_date`

katelovescode (changes requested):
Looks really great, just a couple notes!

spec/models/child_spec.rb
wittejm: I think deleted entries should be included in any scope including `active`, unless the `not_deleted` scope is explicit. One alternative would be to adjust `active` scope to be `where(active: true, deleted: false)` , which seems not great, or to more generally apply `not_deleted` as a default scope everywhere and provide an inversion `include_deleted`. I think the hassle of adding `not_deleted` everywhere is worth it for avoiding the possible confusion. Also it looks like default scopes are frowned upon: https://piechowski.io/post/why-is-default-scope-bad-rails/

spec/models/child_spec.rb
wittejm: Some thing here: inactive and deleted entries should still be present IMO.

wittejm:
Good idea re new validation tests. I discovered that 

`validates :last_active_date, date_param: true, allow_nil: true`

actually passes on an invalid string, because the string is quietly cast to a `nil`. https://stackoverflow.com/a/35553281

So I tweaked the validation rule to check for nil before the type cast. 

My comments above address your other suggested changed.

app/models/child.rb
```diff
@@ -20,10 +20,20 @@ class Child < UuidApplicationRecord
...
   validates :full_name, uniqueness: { scope: %i[date_of_birth business_id] }, unless: -> { errors }
 
+  REASONS = %w[
+    no_longer_in_my_care
+    no_longer_recieving_subsidies
+    other
+  ].freeze
+
+  validates :inactive_reason, inclusion: { in: REASONS }, allow_nil: true
+  validates :last_active_date, date_param: true, unless: proc { |c| c.last_active_date_before_type_cast.nil? }
```
katelovescode, line 30: TIL. 😳😳😳 This is awesome.

katelovescode (approved):

Issue #606, opened 2020-11-25
Write a rake task to get NE subsidy rules into the database 


csprayregen:
📋 Summary
Write a rake task to get NE subsidy rules into the database.

💻 Implementation Details
Acceptance criteria:
- [x] Tables, services, and associations related to to "subsidy rules" are renamed "rates" (including the SubsidyRuleAssociator)
- [x] A CSV parser is built, using patterns established in existing services (or improvements and refactors!) that will parse a CSV with the following fields:
    - State (should be used to determine which state table to store extra fields to but does not need to be stored in its own right)
    - County (Rate)
    - Age (Rate - this is `max_age`)
    - Rate type (NebraskaRate)
    - QRIS enhancement threshold (NebraskaRate)
    - Special needs enhancement threshold (NebraskaRate)
    - Accreditation enhancement threshold (NebraskaRate)
    - Effective on (date) (Rate)
    - Expires on (date) (Rate - optional)
    - License Type
- [x] These fields are stored, as appropriate, on the Rate table and the NebraskaRate table as noted above

Note: there's room here for data model changes; maybe the parent Rate table doesn't make sense - would love some thoughts on this

UPDATE: The model has changed, these just live in the Rate table.

🧐 Why tho
So that our app can correctly calculate predicted subsidy revenue based on state rules

wittejm:
Currently works on a local file

Todo:
- Pull the file from S3 instead of local
- Tests!

dev notes: 
- Minor differences between the expected csv headers and the db columns mean there's this annoying `to_record_entry` method. Can we change the expected headers to match the table? 
- probably can refactor the rake code a little bit so the pattern of "try a step and log failure then abort" flow is cleaner.

Other changes to consider: 
- Merge the `rates` and `illinois_rates` tables. The idea is to abandon this separation between shared and state-specific rate fields, for all states. 
- Merge the `rate_id` and `nebraska_rate_id` columns in `child_approvals`. This would be polymorphic for the different state rate rules.

PR #1228, opened 2021-05-05
[WIP] Rename tables ahead of Rake import nebraska rates csv

wittejm:

#606

Currently this just renames all the relevant tables and columns with "subsidy rule" to "rate", and the "subsidy ruleable" index to "state rule" -- I couldn't decide on a name I really liked, but "state rate" might also be fine because it's an index to "[State]Rate" tables....


wittejm:
Currently works on a local file

Todo:
- Pull the file from S3 instead of local
- Tests!

dev notes: 
- Minor differences between the expected csv headers and the db columns mean there's this annoying `to_record_entry` method. Can we change the expected headers to match the table? 
- probably can refactor the rake code a little bit so the pattern of "try a step and log failure then abort" flow is cleaner.

Other changes to consider: 
- Merge the `rates` and `illinois_rates` tables. The idea is to abandon this separation between shared and state-specific rate fields, for all states. 
- Merge the `rate_id` and `nebraska_rate_id` columns in `child_approvals`. This would be polymorphic for the different state rate rules.

PR #1235, opened 2021-05-06
Dashboard view shows inactive children but hides deleted children

wittejm:
Continuation of #1189

## 🍂 Before You Submit

* [x] all the things

## 🛷 Deployment

* [x] none
## 🧳 Steps to test

Verify dashboard endpoint responses
## 🕯 Caveats, concerns

There might be other occurrences of a similar issue. But I looked around and found no other uses of the `active` scope in the existing code.


wittejm:
I'm not super confident that I got everything correct here. But this is the only occurrence that I found of `active` scope for child, and I made the minimal changes to tests that were required to pass.

katelovescode (changes requested):
You're right - given that we only have a dashboard view right now, this is the only place it's being used.  For now what I'd like to do is chain `.active.not_deleted` in the blueprints because we won't want to see inactive OR deleted records on the dashboard.

As for tests, I would love to see some added scope checks on the users_spec - `case_list_for_dashboard` shouldn't include those deleted records, nor inactive records.  I think those tests were set up a while ago and scoping was never built into the assertions.

cjhaddad:
> You're right - given that we only have a dashboard view right now, this is the only place it's being used. For now what I'd like to do is chain `.active.not_deleted` in the blueprints because we won't want to see inactive OR deleted records on the dashboard.
> 
> As for tests, I would love to see some added scope checks on the users_spec - `case_list_for_dashboard` shouldn't include those deleted records, nor inactive records. I think those tests were set up a while ago and scoping was never built into the assertions.

Hey Kate! In your comment you say "we won't want to see inactive OR deleted records n the dashboard." My understanding is that we do want to see inactive records in the dashboard, for there will be the ability to reactivate them.

wittejm:
What Candice said sounds right, and it looks that way in the Figma design. I added tests to check that the dashboard includes an inactive record and excludes a deleted one. Let me know if these tests should be in a different style, I just imitated similar stuff in `users_spec`

katelovescode (approved):

Issue #913, opened 2021-02-10
Parse attendances into the db to prep for the app doing revenue calculations 


csprayregen:
## 📋 Summary
The goal is to be able to take data directly from WS's attendance data dump and use it for the revenue calculations, once those are implemented in the app. To prepare for this:
- [ ] Create importer rake task
- [ ] Add logging if the user isn't found
- [ ] Add instructions/buckets for manually uploading WS's attendance data to S3
- [ ] Parse the attendance data and make sure to find/overwrite existing records with the same attendance ID so we don't make dupes

## Implementation Details
Open Questions/Out of Scope:
- ~~Can WS add an attendance ID to the table, i.e. their unique ID so we can match on that?~~
- Can we get the attendances from a public CSV?
- Can we scrape the website with the password access to get this information rather than manually downloading it?

katelovescode:
@csprayregen this is on hold while we wait for updates from Wonderschool; we're hoping we can pull the child info from a CSV link automatically but if not we can scrape the website.  For now I think the formulas are most important so I'm going to move on to those and we'll keep doing manual attendance upload.

PR #1252, opened 2021-05-14
[PIE-913] Updates the attendance processor and associated logic

katelovescode:
Updates the attendance processor and associated logic; removes null constraint from check out time, updates the time param validator to use default error handling, adds missing field spec runs for dashboard and attendance, adds wonderschool ID to attendances, adds safe operators to active_child_approval, removes 0 validation on total_time_in_care

#913 

## 🍂 Before You Submit

* [x] Did you write tests?

## 🛷 Deployment

* [x] Schema Migrations

## 🧵 Steps to set up locally

```
bundle exec rails db:migrate
```

## 🧳 Steps to test
correct records:
- create two S3 buckets (one for uploading, one for archiving) if you haven't already and configure your env. vars to use those buckets for the attendance s3 bucket and archive
- put a correctly formatted attendance csv in the uploading S3 bucket (see `spec/fixtures/files/wonderschool_necc_attendance_data.csv` for example) with records for children in your db - the child should have a `wonderschool_id` that matches the `child_id` in the spreadsheet, and belong to a business where the `name` matches the `school_name` in the spreadsheet
- run `bundle exec rails read_wonderschool_necc_attendances`
- [ ] your attendances should be created correctly
- [ ] your uploaded CSV should be archived to the other bucket

invalid format:
- add an invalid format csv to the uploading S3 bucket (see `spec/fixtures/files/wonderschool_necc_attendance_data_invalid_format.csv` for example)
- run `bundle exec rails read_wonderschool_necc_attendances`
- [ ] a `failed_records` csv should be uploaded to the archive s3
- [ ] your original csv should not be archived

missing field:
- add a csv that is missing a required field to the uploading S3 bucket (see `spec/fixtures/files/wonderschool_necc_attendance_data_missing_field.csv` for example)
- run `bundle exec rails read_wonderschool_necc_attendances`
- [ ] a `failed_records` csv should be uploaded to the archive s3
- [ ] your original csv should not be archived

## 🕯 Caveats, concerns
This will have merge conflicts w/ @wittejm 's ongoing PR but I can handle those after his is merged


katelovescode:
@csprayregen for QA on this, it's probably best if we pair since the data and DB on staging will require some setup, and I can walk you through the process.

PR #1254, opened 2021-05-16
Rename subsidy_rule to rate

wittejm:
* and rename `subsidy_ruleable` index to `state_rule`.

Should have filed this as a regular PR earlier rather than glom it into that big WIP branch. 
  
## 🍂 Before You Submit

* [x] updated corresponding tests

Just renaming things, no functional changes


wittejm:
Ok well I don't know why the backend tests failed here the first time. I tested again locally, changed nothing, forced-pushed for good measure, and now it passes, hooray. @katelovescode

app/models/illinois_rate.rb
```diff
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
 # Subsidy rules that apply for Illinois
-class IllinoisSubsidyRule < UuidApplicationRecord
-  has_one :subsidy_rule, as: :subsidy_ruleable, dependent: :restrict_with_error
+class IllinoisRate < UuidApplicationRecord
+  has_one :rate, as: :state_rule, dependent: :restrict_with_error
```
katelovescode, line 5:
```suggestion
  has_one :rate, as: :rateable, dependent: :restrict_with_error
```

Couple reasons for this suggestion:
- it's "standard" syntax for polymorphics
- the state rule *is* the IllinoisRate; the polymorphic rate table record (which table will be deprecated eventually) is the generalized rate info that has nothing to do with the state.

katelovescode (changes requested):

wittejm:
Renamed, rebased, and re-ran `rake prep`

katelovescode (approved):

PR #1266, opened 2021-05-20
Just re-add the new files for the Nebrask Rates model and rake task.

wittejm:
This was essentially cherry-picked from other WIP, so some finagling is required. Also it doesn't run as-is because the changes are extracted away from the "subsidy_rule" model names refactor #1254. Merge that and rebase this branch on that commit before trying to reconcile anything else.

It may or may not actually save dev time!

Todo to complete issue #606:
 * Test and apply any need refactoring/fixes.
 * Write tests
 * Attach to S3. this currently pulls a filename from the command line argument provided to the rake task.

I still have a single unit test for this that I co-wrote with Kate, will clean that up and push as well when I can.


db/schema.rb
```diff
@@ -10,7 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
+<<<<<<< HEAD
```
wittejm, line 13: Woops didn't mean to commit this useless.

wittejm:
I also didn't mean to delete the child_approval and child_approval spec files, not sure how that got into this commit.

PR #1298, opened 2021-05-29
[PIE-714] retrieve hours from api

katelovescode:
closes #714 - this creates a Nebraska Hours calculator (which includes rounding logic), makes sure that when attendances are created they *must* have a child_approval and their total_time_in_care is set to Nebraska expectations (see `app/models/attendance.rb:35` for NE logic.  This PR also adds a "Time of Day" library helper that lets us store times of day in the db like "09:30PM" without a date attached.

## 🍂 Before You Submit

* [x] Did you write tests?

## 🛷 Deployment

* [X] Schema Migrations
* [X] Dependencies

## 🧵 Steps to set up locally

`bundle exec rails db:migrate`
`bundle install`

## 🧳 Steps to test
There's a convenient rake task you can run using `bundle exec rails test714:kids_and_schedules` which will populate one of the NE kids with a schedule, and one without, and give them each an Attendance with a nil `check_out` time.  There are some `puts` statements so when you run the task, it'll tell you which kid will have no hours of attendance (because they defaulted to 8 which would be a full day), and which kid will have a 5-hour attendance (dictated by their schedule for a particular date).  The date used in the rake task is May 3rd, 2021; when it comes time for QA I'll change the date for June so Chelsea can check.

There's also another rake task `bundle exec rails test714:feature_flag` which will help with testing the feature flag that I've built for the live algorithms.  You can set a new environment variable in your `.env`: `FF_LIVE_ALGORITHMS=true` and it will display the calculated hours.  Set it to `false` or remove it, and it'll use the TemporaryNebraskaDashboardCase hours as usual (you'll need to restart the server after changing the environment variable).  The rake task creates a TemporaryNebraskaDashboardCase with 35 hours, and an attendance with 4 hours duration so when you flip between the two environment variables it should be very obvious whether you're looking at the feature flag on or off.

## 🕯 Caveats, concerns
Make sure to look at the tests and verify we're testing the most important behavior.


app/models/child.rb
```diff
@@ -9,6 +9,7 @@ class Child < UuidApplicationRecord
 
   has_many :child_approvals, dependent: :destroy, inverse_of: :child, autosave: true
   has_many :approvals, through: :child_approvals
+  has_many :schedules, dependent: :destroy
```
JonErvin, line 12: Seems straightforward, sets up one to many relationship child to schedules (child has many schedules).  Destroy means destroy schedules when child is destroyed I think?

app/models/schedule.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+# Expected attendance schedules for children
+# used to calculate hours attended compared to hours scheduled
+class Schedule < ApplicationRecord
+  belongs_to :child
```
JonErvin, line 6: Squares with the edit to child above.  Rather than looking up by WS ID in the sheet, this actually attaches it to the child data structure (and whatever its primary key is... I think thats the UUID Application Record?)

katelovescode:
Open question - what happens if a kid attends before school AND after school - do they round on the total for a single day?  Do they round each attendance?  What happens if the attendance rolls into the next day (after midnight), do they total based on the check-in day?

katelovescode:
> Open question - what happens if a kid attends before school AND after school - do they round on the total for a single day? Do they round each attendance? What happens if the attendance rolls into the next day (after midnight), do they total based on the check-in day?

@JonErvin @csprayregen I looked in the Nebraska model document and the prediction spreadsheet to see if we had any information about multiple attendances on the same day; from what I can tell, each attendance is calculated and rounded up to the nearest 15 minutes individually, so if they have multiple check ins the same day, they each get rounded in the spreadsheet.

It also looks like everything is calculated in minutes from check-in time, and the service date is based on the check-in date, so that should answer my "what happens if it rolls into the next day" question.

Let me know if I'm missing anything with this!

csprayregen:
I'll let @JonErvin confirm if this all seems right. Let me know if you'd like me to create a ticket to research this question or refine this formula. Thanks all!

app/models/child.rb
katelovescode: Yep!  Exactly that!

app/models/schedule.rb
katelovescode: Yep!

app/models/attendance.rb
```diff
@@ -23,7 +23,21 @@ class Attendance < UuidApplicationRecord
...
+                                check_out - check_in
+                              elsif child_approval.child.state == 'NE'
+                                calculate_from_schedule
+                              else
+                                0.seconds
+                              end
+  end
+
+  def calculate_from_schedule
+    child_schedule ? child_schedule.end_time.on(check_in.to_date) - check_in : 8.hours
```
JonErvin, line 36: @katelovescode Is this using the chlid's end time in the schedule to fill-in a missing end time?  If so, this could work, but currently we've assumed the total scheduled number of hours rather than using real check-in and scheduled check-out.

PR #1338, opened 2021-06-17
[PIE-698] Adds full_day calculations for Nebraska, based on feature flag

### LINKED ISSUE #698: I can retrieve Full days from the API for the NE dashboard
The API returns the Full days category for the Nebraska dashboard

🧶 Acceptance Criteria
- [x] Ticket #697  has the necessary query for this data documented
- [ ] This passes test cases (to be finalized)
- [ ] This matches formulas laid out by the data team (to be finalized)

🥧 Additional context
Figma link for reference of initial implementation of this query on the front-end: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=3579%3A1

katelovescode:
closes #698 - this creates a Nebraska Full Days calculator based on the P4P Nebraska Model and the calculations being done in the prediction algorithm spreadsheet.  It assumes that any attendance greater than or equal to 5 hours and 45 minutes counts as a full day, unless it is greater than or equal to 14 hours and 45 minutes, at which point it becomes two full days

## 🍂 Before You Submit

* [x] Did you write tests?

## 🧳 Steps to test
There's a convenient rake task you can run using `bundle exec rails test698:kids_and_schedules` which will populate one of the NE kids with a schedule of 7 hours, and one without a schedule at all, and give them each an Attendance with a nil `check_out` time.  The date used in the rake task is June 3rd, 2021; if QA happens in July, I'll change the date before we get there.  Both kids should have 1 full day in their June dashboard when viewed after running this task, with the live algorithms environment variable set to `true` - set the following in your `.env`: `FF_LIVE_ALGORITHMS=true`

## 🕯 Caveats, concerns
Make sure to look at the tests and verify we're testing the most important behavior.

### DIFF
--- a/app/services/nebraska_full_days_calculator.rb
+++ b/app/services/nebraska_full_days_calculator.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# Service to calculate full days used in Nebraska by specific kids
+class NebraskaFullDaysCalculator
+  def initialize(child, filter_date)
+    @child = child
+    @filter_date = filter_date
+  end
+
+  def call
+    calculate_full_days
+  end
+
+  private
+
+  def calculate_full_days
+    @child.attendances.for_month(@filter_date).reduce(0) do |sum, attendance|
+      sum + calculate_full_days_based_on_duration(attendance.total_time_in_care)
+    end
+  end
+
+  def calculate_full_days_based_on_duration(duration)
+    if duration > (5.hours + 45.minutes)
+      1
+    else
+      0
+    end
+  end
+end
--- a/app/services/nebraska_hours_calculator.rb
+++ b/app/services/nebraska_hours_calculator.rb
@@ -20,13 +20,13 @@ class NebraskaHoursCalculator
   end
 
   def round_hourly_to_quarters(duration)
-    adjusted_duration = if duration < (5.hours + 45.minutes)
+    adjusted_duration = if duration <= (5.hours + 45.minutes)
                           duration
-                        elsif duration > 10.hours && duration < (14.hours + 45.minutes)
+                        elsif duration > 10.hours && duration <= (14.hours + 45.minutes)
                           duration - 10.hours
                         else
                           0.minutes
                         end
-    (adjusted_duration.in_minutes / 60 * 4).round / 4.0
+    (adjusted_duration.in_minutes / 15.0).ceil * 15 / 60.0
   end
 end
--- a/app/services/nebraska_hours_calculator.rb
+++ b/app/services/nebraska_hours_calculator.rb
@@ -20,13 +20,18 @@ class NebraskaHoursCalculator
   end
 
   def round_hourly_to_quarters(duration)
-    adjusted_duration = if duration <= (5.hours + 45.minutes)
-                          duration
-                        elsif duration > 10.hours && duration <= (14.hours + 45.minutes)
-                          duration - 10.hours
-                        else
-                          0.minutes
-                        end
-    (adjusted_duration.in_minutes / 15.0).ceil * 15 / 60.0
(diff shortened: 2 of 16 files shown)


app/services/nebraska_full_days_calculator.rb
```diff
@@ -0,0 +1,31 @@
...
+  private
+
+  def calculate_full_days
+    @child.attendances.for_month(@filter_date).reduce(0) do |sum, attendance|
+      sum + calculate_full_days_based_on_duration(attendance.total_time_in_care)
+    end
+  end
+
+  def calculate_full_days_based_on_duration(duration)
+    if duration >= (14.hours + 45.minutes)
```
JonErvin, line 23: This logic may be updated per Slack question on July 7th (updated guidance from NE DHS was more clear)

app/services/nebraska_full_days_calculator.rb
```diff
@@ -0,0 +1,31 @@
...
+      2
+    elsif duration >= (5.hours + 45.minutes)
```
JonErvin, line 25: To discuss: Right now we have this 5.hours + 45.minutes parameter in two places (here + `hours_calculator`).  From a DRY perspective, should these services be combined to return two outputs (hours and days)?  the child model could pull the correct output in `nebraska_hours` and `nebraska_full_days` respectively.

katelovescode: Yeah this is a good point, I like this direction.  Let me see if I can pull that logic out.

katelovescode: On further investigation I want to refactor our service objects into procedural programming so I'm gonna skip this for now.

app/services/nebraska_full_days_calculator.rb
```diff
@@ -0,0 +1,29 @@
...
+  private
+
+  def calculate_full_days
+    @child.attendances.for_month(@filter_date).reduce(0) do |sum, attendance|
+      sum + calculate_full_days_based_on_duration(attendance.total_time_in_care)
+    end
+  end
+
+  def calculate_full_days_based_on_duration(duration)
+    if duration > (5.hours + 45.minutes)
```
katelovescode, line 23: @JonErvin this updates the logic based on the new information about more than 24-hours care

app/services/nebraska_hours_calculator.rb
```diff
@@ -20,13 +20,18 @@ def calculate_hours
...
-    (adjusted_duration.in_minutes / 15.0).ceil * 15 / 60.0
+    (adjusted_duration(duration).in_minutes / 15.0).ceil * 15 / 60.0
+  end
+
+  def adjusted_duration(duration)
+    if duration <= (5.hours + 45.minutes)
+      duration
+    elsif duration > 10.hours && duration <= (14.hours + 45.minutes)
+      duration - 10.hours
+    elsif duration > 24.hours
```
katelovescode, line 31: @JonErvin this updates the logic based on the new information about more than 24-hours care

katelovescode:
This PR intermingles some of the logic for the new math and some rebasing commits - the important part is the math.

app/services/nebraska_hours_calculator.rb
```diff
@@ -20,13 +20,18 @@ def calculate_hours
...
-    (adjusted_duration.in_minutes / 15.0).ceil * 15 / 60.0
+    (adjusted_duration(duration).in_minutes / 15.0).ceil * 15 / 60.0
+  end
+
+  def adjusted_duration(duration)
+    if duration <= (5.hours + 45.minutes)
+      duration
+    elsif duration > 10.hours && duration <= 18.hours
+      duration - 10.hours
+    elsif duration > 18.hours
```
katelovescode, line 31: @JonErvin New 18 hour logic

JonErvin (approved):

PR #1426, opened 2021-07-21
[PIE-1385] Adds validation to attendance to make sure check_out is after check_in

### LINKED ISSUE #1385: Validate the duration isn't negative on the backend
## 📋 Summary
The front-end is going to handle check-in and check-out validation but we need matched validation on the backend

- [X] Nebraska
- [X] Illinois

## 💻 Implementation Details

Add a "check out must be later than check in" validation for attendances

## 🧐 Why tho
So other integrations can also be valid!

katelovescode:

Fixes #1385 by adding a validation to the attendance model to make sure check_out > check_in

## 🍂 Before You Submit

* [ ] Did you write tests?

## 🧳 Steps to test
Run the attendance_batches endpoint and try and add an attendance with a check_out time before the check_in time.  You should get an error.

### DIFF
--- a/app/models/attendance.rb
+++ b/app/models/attendance.rb
@@ -12,6 +12,7 @@ class Attendance < UuidApplicationRecord
 
   validates :check_in, time_param: true, presence: true
   validates :check_out, time_param: true, unless: proc { |attendance| attendance.check_out_before_type_cast.nil? }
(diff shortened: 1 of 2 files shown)


cjhaddad (approved):

cjhaddad:
tested and approved! 🥳

PR #1457, opened 2021-08-02
Refactoring CSV Parsing and adding some test coverage

katelovescode:

No ticket, needed to refactor CSV parsing and add test coverage

## 🍂 Before You Submit

* [X] Did you write tests?

## 🧵 Steps to set up locally

S3:
- you'll need AWS keys to test the Onboarding and Dashboard importer, and you'll need to make sure CSVs with correct data are in the right buckets (I recommend making an Onboarding spreadsheet, onboard the kids, and then run dashboards for each of them)

Not-S3
- Attendances come from wonderschool - get the URL from Kate

## 🧳 Steps to test

- run the tasks for in wonderschool_tasks.rake
- They should succeed and not stop for errors; you can test this by putting invalid csvs in there, or deleting one of your onboarded kids, etc.
- Data should show up in Appsignal for the dev environment for actual errors; skipped attendances (because the child isn't in our system) should not show up in Appsignal

## 🕯 Caveats, concerns
Manually testing isn't super important here, Chelsea will QA, so code 👁️ are fine


spec/lib/appsignal_reporting_spec.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.xdescribe AppsignalReporting do
```
minaslater, line 5: Did you mean to disable this group of specs?

minaslater (approved):
Looks good. Saw an `xdescribe` tho, unsure if it's intentional.

spec/lib/appsignal_reporting_spec.rb
katelovescode: Yes, I'm considering it a TODO

Issue #1376, opened 2021-07-02
When I enter attendance, I see errors when I make a mistake. 


csprayregen:
## ✨ What new feature do you want to see?
When I enter my attendance (as detailed in #1125), I see errors as relevant to my entrance. 

## 🇺🇸 States this change applies to
- [x] Nebraska
- [x] Illinois

## 🧶 Acceptance Criteria
I see the following error states:
- [x] No date entered -->  Ideally, I see this error immediately after I leave the date/absence field. If that doesn't work, it should happen as soon as I start typing in the date or absence field 
- [x] Another question for @cjhaddad @rebeccakarasiktw @katelovescode - how do we want to handle it if someone enters information for a date that's already been entered? This could be an error. This could be an attempt to edit. We'll have more options for this when we get to the View Attendance and Multiple Records per day cards. Do we want to do anything currently? What is default behavior? **RK response: I dont think we will want to do anything currently until we better understand the user behavior. But I think if/when we enhance this feature we would only want a date to be entered one time. Also sorry I accidentally made this into a github issue.**
- [x] If an error is present, the 'Save' button is disabled to avoid making an API request with incomplete data.
- [ ] ~~Time is not valid~~ --> not needed, addressed w/in component
- [ ] ~~Date is invalid~~ --> not needed, addressed w/in component
- [ ] ~~Future dates are invalid~~ --> not needed, addressed w/in component 
- [ ] ~~Neither check in time nor absent status selected (at least one must be selected)~~ --> does not apply since some children may not have attendance.  

## 🖍️ Implementation Details
 [**Figma link**](https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=5194%3A0)

> **NOTE TO ENGINEERS**
> 
> Please ensure that you are running Google Lighthouse and/or WebAIM (Wave) on any UI elements in your pull request to check for accessibility issues.
> 
> If it is out of scope to address those accessibility issues in your pull request (because the issue is in a parent component and needs refactoring to work, for example), **you are responsible for noting those issues in your pull request and writing bug tickets to address the issues - make sure you link the new tickets in your pull request**

## 🥧 Additional context
@rebeccakarasiktw please add figma mockups here when ready

rebeccakarasiktw:
[...]

PR #1464, opened 2021-08-03
Pie 1376/attendance input errors

cjhaddad:
this addresses #1376 and renders an error when a date for the column in not selected and needs to be.
WIP bc disabling of Save logic needs to be tweaked, but putting this PR up now for preliminary review

## 🍂 Before You Submit

* [ ] Did you write tests?

## 🛷 Deployment

* [ ] Data Migrations
* [ ] Schema Migrations
* [ ] Dependencies

### DIFF
--- a/client/src/Attendance/Attendance.js
+++ b/client/src/Attendance/Attendance.js
@@ -26,7 +26,7 @@ export function Attendance() {
   }))
   const [tableData, setTableData] = useState(cases)
   const [isSuccessModalVisible, setSuccessModalVisibile] = useState(false)
-  const [disableSave, setDisableSave] = useState(false)
+  const [errors, setErrors] = useState(false)
   const reduceAttendanceData = data =>
     data.reduce((acc, cv) => {
       return {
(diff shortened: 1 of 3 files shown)


client/src/Attendance/Attendance.js
```diff
@@ -84,15 +92,32 @@ export function Attendance() {
...
-            style={{ width: '8rem ', color: '#004A6E' }}
-          />
-        ),
+        title: () => {
+          const errorIsPresent =
+            Object.values(latestAttendanceData.current).find(
+              row => Object.keys(row[i]).length > 0
+            ) && latestColumnDates.current[i] === ''
+
+          if (errorIsPresent !== latestestDisableSave.current) {
```
katelovescode, line 101:
```suggestion
          if (errorIsPresent !== latestDisableSave.current) {
```

client/src/Attendance/Attendance.js
```diff
@@ -84,15 +92,32 @@ export function Attendance() {
...
+            latestestDisableSave.current = errorIsPresent
```
katelovescode, line 102:
```suggestion
            latestDisableSave.current = errorIsPresent
```

katelovescode (commented):
Minor typo stuff, I'll wait til it's done to do a functionality check

katelovescode:
Couple of issues I found:

#1 - right after the page is loaded the SAVE button is active and is able to send requests to the API, which creates a parameter error in the backend.  The good news is, it doesn't stop the user from continuing, but we should probably prevent them from submitting with no data

save_no_data

#2 - if I get the "must select a date" error, and then pick the date, the Save button doesn't become active again....

save_inactive

...unless I make a change to another field

reactivated_save

From what I can tell these are the only error states I should be testing, based on the AC in the issue, let me know if I missed anything!

katelovescode:
(also your PR still says WIP but I assume this is the one you wanted me to review!)

cjhaddad:
ready for your eyeballs again @katelovescode

katelovescode:
Looks great to me!

katelovescode (approved):

PR #1465, opened 2021-08-03
[PIE-1311] Adds a default schedule to an after_create_commit hook

### LINKED ISSUE #1311: My formulas work even if I'm not a Wonderschool provider
## 📋 Summary
What about the NE formulas will work for non Wonderschool providers? How do we want to handle that?

- [ ] Nebraska
- [ ] Illinois

## 💻 Implementation Details

> **NOTE TO ENGINEERS**
> 
> Please ensure that you are running Google Lighthouse and/or WebAIM (Wave) on any UI elements in your pull request to check for accessibility issues.
> 
> If it is out of scope to address those accessibility issues in your pull request (because the issue is in a parent component and needs refactoring to work, for example), **you are responsible for noting those issues in your pull request and writing bug tickets to address the issues - make sure you link the new tickets in your pull request**

katelovescode:
Adds a default schedule to an after_create_commit hook, updates the onboarding_csv_importer to work with the hook, refines some spec behavior to handle the callback

Closes #1311 

## 🍂 Before You Submit

* [x] Did you write tests?

## 🛷 Deployment

* [x] Schema Migrations

## 🧵 Steps to set up locally

```
bundle exec rails db:migrate
```

## 🧳 Steps to test
Create a child, either through the API call from Postman, or via the console.  The child should have 5 Schedule records attached to their record after it's saved, with 5 days, Mon-Fri, with 8 hours (9am - 5pm).

Then try to create a child with Schedule info
```
$ bundle exec rails console
> schedule = Schedule.new(start_time: "3:00pm", end_time: "6:00pm", effective_on: Time.zone.now - 30.days, weekday: 2)
> child = Child.new(business: Business.first, full_name: "Kid's Name", date_of_birth: "2017-04-03", approvals: [Approval.first])
> child.schedules << schedule
> child.save!
```
The child should have 1 schedule that matches the params you passed, rather than 5 default ones.

Then try to update the child that already has the schedule
```
child.update!(full_name: "New Name")
```
The child should still have 1 schedule that matches the params you passed because this should only happen on create when the child is onboarded.

## 🕯 Caveats, concerns
This will require some finagling if we end up allowing kids to be onboarded with schedules, but the refactor in the onboarding importer in this PR might be some good guidance if we have to do that.


minaslater (commented):
I'm testing the first part of your Steps to Test. Created a new Child with:

```
c = Child.create(date_of_birth: Time.current.strftime('%Y-%m-%d'), full_name: "Mina Slater", business: FactoryBot.create(:business))
```

But `c.schedules` is empty. Maybe I'm missing something.

katelovescode:
> c = Child.create(date_of_birth: Time.current.strftime('%Y-%m-%d'), full_name: "Mina Slater", business: FactoryBot.create(:business))

Your kid isn't actually getting created here, they also need approvals to be saved.  Try this:

```
c = Child.create(date_of_birth: Time.current.strftime('%Y-%m-%d'), full_name: "Mina Slater", business: FactoryBot.create(:business), approvals: [FactoryBot.create(:approval, create_childr
en: false)])
c.reload
c.schedules
```

I'll fix the instructions.

db/migrate/20210803170531_change_unique_index_for_schedules.rb
```diff
@@ -0,0 +1,10 @@
+class ChangeUniqueIndexForSchedules < ActiveRecord::Migration[6.1]
+  def up
+    remove_index :schedules, name: "unique_child_schedules"
+    add_index :schedules, %i[effective_on child_id weekday], unique: true, name: :unique_child_schedules
+  end
+  def down
```
minaslater, line 6: ~Your `down` is doing super similar things to the `up`. 🤔 I would expect the `down` to do the opposites?~

NVM.

minaslater (approved):
👍

PR #1502, opened 2021-08-10
[PIE-712] Earned revenue from dashboard

katelovescode:
## 💅🏼 What issue does this fix?

Closes #712 

## 🍂 Before You Submit

* [X] Did you write tests?
* [X] Did you run `bundle exec rspec` from the root?
* [X] Did you run `bundle exec rubocop` from the root?
* [X] Are your primary keys UUIDs on any new tables?

## 🛷 Deployment

* [X] Schema Migrations

## 🧵 Steps to set up locally

```
bundle exec rails db:migrate
```

## 🧳 Steps to test
- create an accredited hourly Nebraska Rate in the LDDS region with an amount of 5.76
- create a business in one of the LDDS counties with a qris rating of step_five and make the business accredited
- create a child for that business (don't forget to add an active approval as well)
- set that child's active child approval to `special_needs_rate: false`
- create an attendance for that child during the active child approval period - let's say 3 hours and 21 minutes total_time_in_care
- the earned revenue for your created attendance should be 3.5 (total time in care rounded up to nearest 15 mins) * 5.76 (rate amount you set above) * 1.1576 (hardcoded step_five qris bump)

You can run a bunch of other manual tests, turning `special_needs_rate` on (and setting `special_needs_hourly_rate` or `special_needs_daily_rate`), turning accreditation off, setting the business' qris rating to `not_rated`, `step_one`, etc.

The best way to end-to-end test is to take ONE child from our current NE prediction algorithms spreadsheet, replicate them in your local, and replicate their attendances in your local, and then check to see if what you have in your dashboard for that child (after turning on the nebraska live algorithms feature flag) matches what the spreadsheet has for that child.

## 🕯 Caveats, concerns
This is the first time we're including rates, and it's based on several assumptions that are outlined in our #data slack channel, so if any of those are wrong we'll need to iterate.

Here are my "steps to test" for one of the kids in our prod system:

```
Steps to Test:
- reset & reseed db
- child = Child.find_by(full_name: "Rhonan Shaw")

Create Rates:

bundle exec rails nebraska_rates20210813

business changes:
accredited: false
step_two
county: Buffalo

Child record changes:
- change birthdate to match Client #89646689
- set enrolled_in_school to false

Approval record changes:
- Effective on: 2021-03-01
- Expires on: 2022-02-28

ChildApproval record changes:
- Special Needs Rate = false
- Special Needs Daily Rate = nil
- Special Needs Hourly Rate = nil
- authorized weekly hours = 45

[...]

PR #1547, opened 2021-08-24
[PIE-713] absences from dashboard

### LINKED ISSUE #713: I can retrieve Absences from the API for the NE dashboard
The API returns the Absences category for the Nebraska dashboard

🧶 Acceptance Criteria
- [ ]  This matches formulas laid out by the data team
- [ ]  This passes test cases
- [ ] The dashboard displays "X of 5" because Nebraska limits the number of absences allowed to 5 (decided 8/27)

🥧 Additional context
Figma link for reference of initial implementation of this query on the front-end: https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=3579%3A1

# Implementation Details

- [ ] make sure the calculations don't count more than 5 absences, but return as many absences as have been entered

## Open questions
- what should we do if someone tries to enter a 6th absence with the add attendance behavior Candice has been working on? cc @csprayregen @rebeccakarasiktw. 
- @csprayregen we should talk about the covid-related absence thing, maybe wonderschool can implement a checkbox

katelovescode:

Fixes #713 

This adds dynamic absences to the dashboard behind a feature flag, and updates the revenue calculations to ensure no more than 5 absences are counted towards revenue numbers each month.

## 🍂 Before You Submit

* [X] Did you write tests?

## 🧵 Steps to set up locally

Same as the former PRs starting with #1502 - start adding rates and attendances and let's check calculations of revenue

## 🧳 Steps to test
- `absences` on the dashboard should now say `X of 5`


app/models/child.rb
```diff
@@ -147,13 +147,20 @@ def nebraska_estimated_revenue(filter_date)
...
+    [(absence_revenue(filter_date) + attendance_revenue(filter_date) - active_nebraska_approval_amount(filter_date)&.family_fee.to_f), 0.0].max
+  end
+
+  def attendance_revenue(filter_date)
+    attendances.non_absences.for_month(filter_date).pluck(:earned_revenue).sum
+  end
+
+  def absence_revenue(filter_date)
+    absences, covid_absences = attendances.absences.for_month(filter_date).partition { |absence| absence.absence == 'absence' }
+    absences.take(5).pluck(:earned_revenue).sum + covid_absences.pluck(:earned_revenue).sum # only five absences are allowed per month in Nebraska
```
JonErvin, line 159: @katelovescode How are we selecting which 5 to take?  If we don't do this already, can we sort by descending earned_revenue and pluck those 5?  JT had some creative logic to make this happen in the spreadsheet model.

app/models/child.rb
```diff
@@ -68,6 +68,14 @@ def attendances
     Attendance.joins(:child_approval).where(child_approvals: { child: self })
   end
 
+  def absences(filter_date)
```
abdulfataiaka, line 71: [QUESTION] This method looks like it returns a presentation `x of n` required on the UI. Is there an intention of being able to reuse it for non UI operations?

app/models/child.rb
```diff
@@ -68,6 +68,14 @@ def attendances
...
+    if Rails.application.config.ff_ne_live_algorithms
+      "#{attendances.for_month(filter_date).absences.length} of 5"
```
abdulfataiaka, line 73: [QUESTION] Why are we hardcoding `5` here?

PR #1568, opened 2021-08-28
No ticket/fix family fee calc and add test

katelovescode:

Fixes the family fee bug, and adds an additional attendance test

## 🍂 Before You Submit

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,11 @@
-# README
+| CI | Analytics |
+| :-: | :-: |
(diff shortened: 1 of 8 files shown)


JonErvin (commented):
Hey this all looks great!  Thanks!

I have one thought on a useful test, but I'm not sure where/how I would implement this.  When we were putting together the spreadsheet version, we tried to raise an error any time the full copay wasn't allocated to the children appropriately.  The pseudo code is something like

`For each approval:
   Expect approval.copay == SUM approval.children.nebraska_family_fee(today) `

This should always be true, regardless of how we allocate the fee.  If it's not, we hit a weird case where our math is wrong.  Thoughts on if this is possible and where something like this could go?

Issue #1447, opened 2021-07-27
Flaky attendances spec


katelovescode:
## 📋 Summary

This one is weird; I can't replicate it locally with any of the seed numbers that have failed so far.

![Screen Shot 2021-07-27 at 3 16 14 PM](https://user-images.githubusercontent.com/8364647/127228322-b9bbb4c3-7084-45fa-803f-78dcff517f0a.png)

## 🇺🇸 States this change applies to
- [ ] Nebraska
- [ ] Illinois

## 💻 Implementation Details

> **NOTE TO ENGINEERS**
> 
> Please ensure that you are running Google Lighthouse and/or WebAIM (Wave) on any UI elements in your pull request to check for accessibility issues.
> 
> If it is out of scope to address those accessibility issues in your pull request (because the issue is in a parent component and needs refactoring to work, for example), **you are responsible for noting those issues in your pull request and writing bug tickets to address the issues - make sure you link the new tickets in your pull request**

katelovescode:
It's back!  See Actions run today ☹️

katelovescode:
@abdulfataiaka if you're at all interested in looking at this, I have a new theory.  I think this only fails *near the end of the month* so it's possible the date/time math around this test setup is broken (or that the behavior itself breaks near the end of the month 😱 ).  Something to consider: all of the logic should be in the user's timezone *except* the fact that we store attendances in UTC for accuracy.  If you've got opinions about timezones, though, I'd love to hear them :)

abdulfataiaka:
Sure, I can take this up

abdulfataiaka:
Some previous failures
- https://github.com/pieforproviders/pieforproviders/runs/3450629917?check_suite_focus=true
- https://github.com/pieforproviders/pieforproviders/runs/3450577396?check_suite_focus=true
- https://github.com/pieforproviders/pieforproviders/runs/3448518458?check_suite_focus=true
- https://github.com/pieforproviders/pieforproviders/runs/3448498656?check_suite_focus=true
- https://github.com/pieforproviders/pieforproviders/runs/3448497382?check_suite_focus=true

abdulfataiaka:
Based on investigation, this failure is expected to only happen on Fridays and Saturdays, each with a different reason

**Failures on Fridays**

[...]

PR #1606, opened 2021-09-09
Update Attendance Spec

abdulfataiaka:

https://github.com/pieforproviders/pieforproviders/issues/1447

Update `attendance_spec` to fix flakiness it tests runs on CI

## 🍂 Before You Submit

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,10 +62,8 @@ end
 group :test do
   gem 'database_cleaner'
   gem 'json-schema'
-  gem 'mocha'
   gem 'shoulda-matchers'
   gem 'simplecov', require: false
-  gem 'timecop'
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
--- a/Gemfile
+++ b/Gemfile
@@ -62,11 +62,11 @@ end
 group :test do
   gem 'database_cleaner'
   gem 'json-schema'
+  gem 'mocha'
   gem 'shoulda-matchers'
   gem 'simplecov', require: false
+  gem 'timecop'
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
 gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
-
-gem 'mocha', group: :test
--- a/spec/requests/api/v1/attendances_spec.rb
+++ b/spec/requests/api/v1/attendances_spec.rb
@@ -7,9 +7,9 @@ RSpec.describe 'Api::V1::Attendances', type: :request do
   let!(:business)       { create(:business, user: logged_in_user) }
   let!(:child)          { create(:child, business: business) }
 
-  let!(:week_mid_date)   { Time.new(2021, 9, 8) }    # Wed, 8th Sept, 2021
-  let!(:week_start_date) { week_mid_date - 3.days }  # Sun, 5th Sept, 2021
-  let!(:week_end_date)   { week_mid_date + 3.days }  # Sat, 11th Sept, 2021
+  let!(:week_mid_date)   { Time.new(2021, 9, 8).in_time_zone }  # Wed, 8th Sept, 2021
+  let!(:week_start_date) { week_mid_date - 3.days }             # Sun, 5th Sept, 2021
+  let!(:week_end_date)   { week_mid_date + 3.days }             # Sat, 11th Sept, 2021
 
   let!(:two_weeks_ago_week_mid_date)   { week_mid_date - 2.weeks }   # Wed, 25th Aug, 2021
   let!(:two_weeks_ago_week_start_date) { week_start_date - 2.weeks } # Sun, 22th Aug, 2021
--- a/spec/requests/api/v1/attendances_spec.rb
+++ b/spec/requests/api/v1/attendances_spec.rb
@@ -4,16 +4,32 @@ require 'rails_helper'
 
 RSpec.describe 'Api::V1::Attendances', type: :request do
   let!(:logged_in_user) { create(:confirmed_user) }
-  let!(:business) { create(:business, user: logged_in_user) }
-  let!(:child) { create(:child, business: business) }
+  let!(:business)       { create(:business, user: logged_in_user) }
+  let!(:child)          { create(:child, business: business) }
+
+  let!(:week_mid_date)   { Time.new(2021, 9, 8) }    # Wed, 8th Sept, 2021
+  let!(:week_start_date) { week_mid_date - 3.days }  # Sun, 5th Sept, 2021
+  let!(:week_end_date)   { week_mid_date + 3.days }  # Sat, 11th Sept, 2021
+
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -62,11 +62,11 @@ end
 group :test do
   gem 'database_cleaner'
   gem 'json-schema'
+  gem 'mocha'
   gem 'shoulda-matchers'
   gem 'simplecov', require: false
+  gem 'timecop'
```
abdulfataiaka, line 68: Not used yet, but figured it would be useful since we deal with dates and time a lot

Gemfile
```diff
@@ -62,11 +62,11 @@ end
 group :test do
   gem 'database_cleaner'
   gem 'json-schema'
+  gem 'mocha'
```
katelovescode, line 65: Do these changes actually need mocha and timecop?

spec/requests/api/v1/attendances_spec.rb
```diff
@@ -4,24 +4,41 @@
 
 RSpec.describe 'Api::V1::Attendances', type: :request do
   let!(:logged_in_user) { create(:confirmed_user) }
-  let!(:business) { create(:business, user: logged_in_user) }
-  let!(:child) { create(:child, business: business) }
+  let!(:business)       { create(:business, user: logged_in_user) }
```
katelovescode, line 7: Can you double check your rubocop or editor config? Consistently throughout the other specs we're using a single space instead of lining them up; would love to keep it consistent (but feel free to make a ticket to change that style!) - if rubocop doesn't automatically edit this, I'd love to see if it's something we can turn on, just so it's consistent across the app.

spec/requests/api/v1/attendances_spec.rb
```diff
@@ -4,24 +4,41 @@
...
+  let!(:child)          { create(:child, business: business) }
+
+  let!(:week_mid_date)   { Time.new(2021, 9, 8).in_time_zone }  # Wed, 8th Sept, 2021
+  let!(:week_start_date) { week_mid_date - 3.days }             # Sun, 5th Sept, 2021
```
katelovescode, line 11: Instead of adding days, can we use the `.at_beginning_of_week` and `.at_end_of_week helpers` just to stay consistent?

katelovescode (changes requested):
Looks good, just a couple of minor changes!

Gemfile
abdulfataiaka: For Timecop - https://github.com/pieforproviders/pieforproviders/pull/1606#discussion_r705641957

For mocha, I simply moved it up, it was there before. But speaking of it now, maybe we should remove it since it doesn't seem like we are using it

spec/requests/api/v1/attendances_spec.rb
abdulfataiaka: Thanks for pointing that out. I actually don't have a preference, just a habit that leaked out 😄. Will look into that

spec/requests/api/v1/attendances_spec.rb
abdulfataiaka: Make sense

Gemfile
katelovescode: Let's leave timecop out for now - RSpec now has native time travel: https://til.hashrocket.com/posts/846d1b9d55-time-travelling-in-rspecrails and as for now we don't need any kind of `tick` behavior with live updates.

As for mocha, I unless it's required by Cypress, which it might be, let's remove it.  But if it's necessary, leave it in.

abdulfataiaka: Removed mocha since it seems to not be used

spec/requests/api/v1/attendances_spec.rb
abdulfataiaka: We can revisit code styles later

spec/requests/api/v1/attendances_spec.rb
```diff
@@ -6,22 +6,42 @@
...
   end
-  let!(:extra_attendances) { create_list(:attendance, 3, check_in: Faker::Time.between(from: Time.current.at_beginning_of_week(:sunday), to: Time.current)) }
 
   describe 'GET /api/v1/attendances' do
     include_context 'correct api version header'
 
-    before { sign_in logged_in_user }
+    before do
+      sign_in logged_in_user
+      allow(logged_in_user).to receive(:timezone).and_return('UTC')
```
abdulfataiaka, line 39: Part of the flakiness was due to the way the app changes the Time.zone to the logged-in user's timezone. So with RSpec, timezone change happens and make time and date inconsistent with the examples

abdulfataiaka:
This PR is ready for a re-review  cc: @katelovescode

spec/requests/api/v1/attendances_spec.rb
katelovescode: I changed the test environment configuration to set the timezone to UTC - I think if we force the *user* to UTC we might miss some behavior that's due to a user logging in from (say) the Central time zone.  We're explicitly storing attendances in UTC but we want to make sure that if someone logs in at 8am Central, they're seeing the attendances we expect.

Can you rebase from develop, which has the configuration change, and see if you can remove this particular `allow` statement and still get the test to pass?

spec/requests/api/v1/attendances_spec.rb
```diff
@@ -44,10 +64,8 @@
     end
 
     context 'when viewed by an admin' do
-      before do
-        admin = create(:admin)
-        sign_in admin
```
katelovescode, line 49: Does this not break the test?  I would assume we have to use the devise test helpers to sign the admin in in order for them to see the right information - if not, we might have a security hole O_O

katelovescode (changes requested):
Couple more things, this is shaping up to be a great PR!

spec/requests/api/v1/attendances_spec.rb
abdulfataiaka: Not at all, it's in the way RSpec works. Since the wrapping describe block has the line

`before { sign_in logged_in_user }`

then the sign_in action will be executed for each example in each context, so all we need to do is override who the `logged_in_user` is when needed, hence the line

`let!(:logged_in_user) { create(:admin) }`

For the admin context

katelovescode: Oh DUH.  Thank you for reminding me to read with my eyeballs and for comprehension :/

spec/requests/api/v1/attendances_spec.rb
```diff
@@ -2,26 +2,47 @@
 
 require 'rails_helper'
 
+# Since the for_week scope on attendance model uses the Sunday to Friday week range, we should make
```
katelovescode, line 5: 😍

katelovescode (approved):
Spectacular!  Nice work, @abdulfataiaka

PR #1753, opened 2021-10-03
Standardize soft-delete

query-string:
## 🍂 Before You Submit

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies


app/controllers/api/v1/businesses_controller.rb
```diff
@@ -46,7 +46,7 @@ def update
       # DELETE /businesses/:id
       def destroy
         # soft delete
-        @business.update!(active: false)
+        @business.update!(active: false, deleted_at: Date.today)
```
katelovescode, line 49: Good call out here in your caveats, @query-string - do you mind changing this behavior to remove setting active to false?

config/routes.rb
```diff
@@ -111,4 +112,3 @@
 #       letters GET    /                                letter_opener_web/letters#index
 #        letter GET    /:id(/:style)(.:format)          letter_opener_web/letters#show
 #               GET    /:id/attachments/:file(.:format) letter_opener_web/letters#attachment
-# rubocop:enable Layout/LineLength
```
katelovescode, line 114: Was this an accidental removal?  Or does rubocop turn disables on at the end of a file automatically now? 🤔

katelovescode (changes requested):
Just a couple of notes, one requested change and one open question - really great work, thank you for the granular migrations!

config/routes.rb
query-string: Yep, it was applied automatically for some reason. I'm happy to roll it back, but suspect it will be applied again during upcoming migrations anyway...

app/controllers/api/v1/businesses_controller.rb
katelovescode: Oh also, please use Time.current instead!

katelovescode (changes requested):

config/routes.rb
katelovescode: If you could roll it back, that'd be great!

query-string:
@katelovescode ✔️

katelovescode (approved):
Looks great, thank you!

Issue #1144, opened 2021-04-09
I can retrieve "daily units remaining" during the approval period from the API


csprayregen:
## 🖍️ Implementation Details

> **NOTE TO ENGINEERS**
> 
> Please ensure that you are running Google Lighthouse and/or WebAIM (Wave) on any UI elements in your pull request to check for accessibility issues.
> 
> If it is out of scope to address those accessibility issues in your pull request (because the issue is in a parent component and needs refactoring to work, for example), **you are responsible for noting those issues in your pull request and writing bug tickets to address the issues - make sure you link the new tickets in your pull request**

katelovescode:
@csprayregen is this the "of x" part of the full days column?  Or is this a different column that isn't implemented yet?

csprayregen:
@katelovescode on the dashboard itself, we've changed this field to just say full days and full hours used. So "5" instead of "5 of 10." Tweaking the name of the card to make that clearer

csprayregen:
@katelovescode we haven't implemented this card yet in design or data, so I'd de-prioritize for now

PR #1763, opened 2021-10-05
[PIE-1143, PIE-1144] Updates the child blueprint to return hourly and daily units remaining and authorized

### LINKED ISSUE #1143: I can retrieve "hourly units remaining" during the approval period from the API
The API returns the hourly units remaining during the approval period column/field for the Nebraska dashboard

🧶 Acceptance Criteria
- [ ] Ticket #1145  has the necessary query for this data documented
- [ ]  This matches formulas laid out by the data team
- [ ] This passes test cases

🥧 Additional context
Figma link for reference of initial implementation of this query on the front-end (see V2): https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=4538%3A16605

The formulas for this are address in the data ticket #1146

katelovescode:

Closes #1143 and #1144 

This adds four fields to the child blueprint to return hourly and daily units remaining and authorized

## 🍂 Before You Submit

## 🧳 Steps to test

Hit the `/case_list_for_dashboard` endpoint from a nebraska account after authenticating.  You should see 4 new fields in every case that's returned: `hours_remaining`, `hours_authorized`, `full_days_remaining`, `full_days_authorized`.

## 🕯 Caveats, concerns

Note: this is different from what we have done in the past on the blueprints for the API - usually I'd concatenate the strings together to match the Figma in the blueprints.  Instead, I'm returning the raw values and expecting the front end to concatenate them together  cc @cjhaddad - I'm going to be doing this more going forward, returning raw data types rather than stringifying everything, and building less custom-api-endpoint blueprinting.  We'll talk about this more later but just wanted to give you a heads up for building this.

### DIFF
--- a/app/models/attendance.rb
+++ b/app/models/attendance.rb
@@ -92,13 +92,17 @@ class Attendance < UuidApplicationRecord
   def ne_hours
     # TODO: this is super sloppy because this shouldn't be a
     # service class but we haven't refactored these to procedures yet
-    NebraskaHoursCalculator.new(child, check_in).round_hourly_to_quarters(total_time_in_care.seconds)
+    NebraskaHoursCalculator.new(child: child,
+                                date: check_in,
+                                scope: :for_month).round_hourly_to_quarters(total_time_in_care.seconds)
   end
 
   def ne_days
     # TODO: this is super sloppy because this shouldn't be a
     # service class but we haven't refactored these to procedures yet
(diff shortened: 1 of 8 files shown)


cjhaddad:
@katelovescode LGTM! confirmed that these new values are coming through on the frontend! Also, acknowledging your note about returning raw data from the backend going forward -- this sounds good and makes sense to me!

Issue #1642, opened 2021-09-15
Each child's individual payment is saved by hitting the backend API


katelovescode:
## ✨ What new feature do you want to see?

The API needs to accommodate saving a payment that is distributed across multiple children

## 🇺🇸 States this change applies to
- [X] ALL
- [ ] Nebraska
- [ ] Illinois

## 🧶 Acceptance Criteria

- [x] Blocked by #1644
- [ ] A `payments_controller` is created to accept these fields via a POST request
- [ ] A route is created for that controller method
- [ ] The appropriate request specs are created and pass

## 🥧 Implementation Details

add similar tests to:
```
context 'when sent with all required fields'
  it 'creates attendances and returns successful records' do
  context 'when the child has no active approval for that time period'
    it 'returns an error'

context 'when missing required fields on one record'
  it 'returns json errors'

context 'when missing required fields on all records'
  it 'returns json errors'

context "when adding an attendance for a child not in the user's care"
  it 'returns json errors'
```

PR #1764, opened 2021-10-06
dragonbone81/pie 1642/batch payments

dragonbone81:
works on #1642 

## 🍂 Before You Submit

## 🕯 Caveats, concerns
Do we need a controller for getting one payment, or getting child payments or payments by month (not sure what the exact params would be but the UI would be displaying payments so I assume they would be fetched)


config/routes.rb
```diff
@@ -76,6 +77,12 @@
...
+#                                          POST   /api/v1/payments_batches(.:format)                                                                api/v1/payments_batches#create {:format=>:json}
+#                           payments_batch GET    /api/v1/payments_batches/:id(.:format)                                                            api/v1/payments_batches#show {:format=>:json}
+#                                          PATCH  /api/v1/payments_batches/:id(.:format)                                                            api/v1/payments_batches#update {:format=>:json}
+#                                          PUT    /api/v1/payments_batches/:id(.:format)                                                            api/v1/payments_batches#update {:format=>:json}
+#                                          DELETE /api/v1/payments_batches/:id(.:format)                                                            api/v1/payments_batches#destroy {:format=>:json}
 #                              attendances GET    /api/v1/attendances(.:format)                                                                     api/v1/attendances#index {:format=>:json}
```
dragonbone81, line 86: Not sure what to do about these? only batch is implemented

spec/requests/api/v1/payments_batches_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'rails_helper'
```
dragonbone81, line 1: I assume this is no-where near enough tests, but I'm also not too sure what we want to test here?
Maybe @katelovescode can give a couple test cases :)))))

dragonbone81:
I assume I'll need to add more tests and maybe more controllers, but I'm down to merge it into our branch (as long as the basic functionality is more or less there) so that it unblocks other people and the UI can do something.

PR #1904, opened 2021-11-08
Testing travel_back

katelovescode:
## 🍂 Before You Submit

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies


spec/rails_helper.rb
```diff
@@ -84,6 +84,7 @@
   config.include FactoryBot::Syntax::Methods
   config.include ActiveJob::TestHelper
   config.include Devise::Test::IntegrationHelpers, type: :request
+  config.include RSpec::Rails::MinitestLifecycleAdapter
```
pirj, line 87: You don't need to do that, [it is included to the base Rails example group](https://github.com/rspec/rspec-rails/blob/fe95eacb376f7fa558d3ffeeaa9afa6d3110a540/lib/rspec/rails/example/rails_example_group.rb#L9) that is included in all [Rails example groups](https://github.com/rspec/rspec-rails/tree/fe95eacb376f7fa558d3ffeeaa9afa6d3110a540/lib/rspec/rails/example).

katelovescode: Interestingly, this passed on CI but failed on my local - no matter whether this was included or not.  If I wanted to do a custom type, like, say `type: :service` because the logic for some of these tests isn't really a "model", would I want to include this, or somehow wrap that particular type in a way that the lifecycle adapter will kick in? 🤔

katelovescode: Even running the same seed as the successes gives me failures locally.  That is bizarre to me.  I've set my test env timezone to UTC to match the CI server; there's lots of dependence on time/date math in this app, so I would expect flakiness if travel-backs don't work.  But *with* the travel_backs it passes both places.  I am wildly confused.

katelovescode:
Deprioritizing

Issue #1557, opened 2021-08-27
Add API Documentation


katelovescode:
## 📋 Summary

Investigate self-documenting API code, rdoc or something similar.

## 🇺🇸 States this change applies to
- [X] ALL
- [ ] Nebraska
- [ ] Illinois

## 💻 Implementation Details

Work was started on this branch: `PIE-1557/add-dox` and this closed PR: https://github.com/pieforproviders/pieforproviders/pull/1922

abdulfataiaka:
This looks like a good issue to start with in understanding the API endpoints we have

abdulfataiaka:
These are the candidates I will be comparing to choose a better one

https://github.com/ruby/rdoc
https://github.com/zipmark/rspec_api_documentation
https://github.com/Apipie/apipie-rails

abdulfataiaka:
RDoc - https://github.com/ruby/rdoc

PROS
- Majorly for ruby code documentation

CONS
- Not suitable for data serving endpoints documentation
- Requires management of documentation comments along with codes
- Possibilities of comments and codes been out of sync

abdulfataiaka:
#### RAD ( https://github.com/zipmark/rspec_api_documentation )

PROS
- Focuses on request endpoints
- Forces us to write acceptance tests
- Make acceptance tests more declarative and readable by using the DSL it provides us
- Support for viewers (ex. apitome) to get a presentation different from the default
- Serves the `/docs` part by default
- Support for different generated documentation files formats

CONS
-

abdulfataiaka:
#### API Pie Rails ( https://github.com/Apipie/apipie-rails )

PROS
- Provides a DSL for declaratively annotating controller action methods
- Support for generating docs from spec examples

CONS
- Uncommon tool with a little bit of learning curve

katelovescode:
@abdulfataiaka one last thing I'd like you to investigate: can you look into how we can use Postman to do API documentation, if it's automated, etc?  I've used rswag before, which uses its own dsl in request specs to document the API, and the problem is that the DSL can be a little challenging, especially for onboarding new devs; I've ended up pulling all of those specs out and replacing them with vanilla RSpec request specs.  So I'm hesitant to go back down that path again with the DSL specs.  Any other options we have?

abdulfataiaka:
Your concern make sense to me, I will take a look into Postman

abdulfataiaka:
[...]

PR #1922, opened 2021-11-12
Add base implementation for Dox

abdulfataiaka:

https://github.com/pieforproviders/pieforproviders/issues/1557

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies

No migrations but requires `redoc-cli` to be installed locally and the addition of the `dox` gem

## 🧵 Steps to set up locally

Run bundle to install new gems and also install the `redoc-cli`
```

$ yarn install-all

$ bundle

```

Run rake task to generate documentation into `public/api_docs/v1` folder
```

$ rails api_docs:generate

# OR

$ bundle exec rails api_docs:generate

```

Open the documentations in the browser
```

http://localhost:3001/api_docs?format=json&version=v1

http://localhost:3001/api_docs?format=html&version=v1

```

## 🕯 Caveats, concerns

This implementation has not been integrated into the build process, I would recommend we have a separate issue to handle that, to ensure much care is taken to prevent breaking production


katelovescode (changes requested):
Couple of notes:

You don't have to install the cli globally, instead run:
```
yarn add redoc-cli --dev
```
 and it will add it to the package.json in the root of the project (then commit it so others can use it!)  Anyone who's going to use it will then just need to run `yarn install` from the root.

Also I couldn't get the html view to work:
Screen Shot 2021-11-15 at 3 07 52 PM

abdulfataiaka:
> Couple of notes:
> 
> You don't have to install the cli globally, instead run:
> 
> ```
> yarn add redoc-cli --dev
> ```
> 
> and it will add it to the package.json in the root of the project (then commit it so others can use it!) Anyone who's going to use it will then just need to run `yarn install` from the root.
> 
> Also I couldn't get the html view to work: Screen Shot 2021-11-15 at 3 07 52 PM

Ensure to run the rake task -> `rails api_docs:generate` first

abdulfataiaka:
@katelovescode Comments and suggestions have been addressed

katelovescode:
> Ensure to run the rake task -> `rails api_docs:generate` first

I did before, and I just tried again, and still no HTML view, unfortunately - what command are you running to start your rails server?

I've tried `bundle exec rails s` and `bundle exec rails start` - and in both cases I tried port 3000 and 3001 and no such luck.

What am I missing? 😭

abdulfataiaka:
That's weird, Lets talk more about it after our check-in session today

abdulfataiaka:
I simply do `foreman start` with the API started at the port `3001`. Could it be that your configuration runs the API at a different port?

katelovescode:
We're going to deprioritize this for now.

Issue #1090, opened 2021-03-26
Make date filtering visible in Nebraska


csprayregen:
## 📋 Summary
After we transition from the low code to the fully automated prototype, enable date filtering features on the NE dashboard to match what is already implemented in IL. 

## 💻 Implementation Details

Before releasing this ticket to users, we should also complete #1720, to minimize their confusion.

Additional enhancement to clarify the user experience are also ticketed but can come after we release date filtering to users. Those are covered in: #1730, #1729, and #1728

> **NOTE TO ENGINEERS**
> 
> Please ensure that you are running Google Lighthouse and/or WebAIM (Wave) on any UI elements in your pull request to check for accessibility issues.
> 
> If it is out of scope to address those accessibility issues in your pull request (because the issue is in a parent component and needs refactoring to work, for example), **you are responsible for noting those issues in your pull request and writing bug tickets to address the issues - make sure you link the new tickets in your pull request**

katelovescode:
@csprayregen and @rebeccakarasiktw is this still blocked by design/needs more info?

csprayregen:
@katelovescode I don't think it's blocked. There are some open questions above, but I'm hoping we can release this and and then address those in follow up tickets. Would love for you two to review those questions and make sure you agree

katelovescode:
>  When I look at previous months, the hours used per week column will not appear currently. We believe it will show the current date, regardless of date filtering settings, which is confusing for the user. OPEN QUESTION: should the column disappear as a whole? Just be blank?

As far as I remember, the decision we made was that it will always show the *last week of the month* in prior months without any way to change it.

> If I look at a previous month for which there is no family fee information, what does the user see? Assume family fee is $0 for that month? Say unknown or n/a?

I think it defaults to $0

csprayregen:
@katelovescode please go through the open questions above and document what the default or current state is, i.e. if we turn on date filtering, what will the user see? We can then determine next steps

katelovescode:
Answers to the above cc @csprayregen and @rebeccakarasiktw 

>  When I look at previous months, the hours used per week column will not appear currently. We believe it will show the current date, regardless of date filtering settings, which is confusing for the user. OPEN QUESTION: should the column disappear as a whole? Just be blank?

Currently this does show the current week, the above "we believe" statement is correct.

>  If I look at a previous month for which there is no family fee information, what does the user see? Assume family fee is $0 for that month? Say unknown or n/a?

[...]

PR #1935, opened 2021-11-18
[PIE-1720] Makes the weekly hours column conditional

### LINKED ISSUE #1720: When I look at an earlier month, my hours per week column is not rendered for that month
## ✨ What new feature do you want to see?
Once date filtering is turned on (see #1090), we want to avoid user confusion with the hours per week column. The current default is for the column to show the CURRENT week, regardless of what month the user is viewing. Instead of this default, we should simply not render the column at all when the user is looking at any month besides the current month.  

- [X] ALL
- [ ] Nebraska
- [ ] Illinois

## 🧶 Acceptance Criteria

- [ ] When I look at any month besides the current month, the hours per week column does not appear in my dashboard table
- [ ] This works in English and Spanish
- [ ] Check for accessibility

katelovescode:
Makes the weekly hours column conditional on filter date being undefined or matching today's month

Closes #1720 

## 🍂 Before You Submit

## 🧳 Steps to test

Currently this can't be tested in isolation; #1090 needs to be finished first


cjhaddad (approved):
LGTM

PR #1992, opened 2021-12-01
No ticket/bug from dropdown

katelovescode:

#1090 surfaced this issue

## 🍂 Before You Submit

## 🧳 Steps to test

- Log into the dashboard using nebraska@test.com and admin@test.com
- Choose a different month in the dropdown
- You should see new data in the attendances and revenue columns and the Hours Attended column should disappear

## 🕯 Caveats, concerns

@cjhaddad please double check my work, I'm rusty at this useState and passing of methods stuff.

### DIFF
--- a/client/src/Attendance/Attendance.js
+++ b/client/src/Attendance/Attendance.js
@@ -120,7 +120,7 @@ export function Attendance() {
                 style={{ width: '8rem ', color: '#004A6E' }}
               />
               {columnErrorIsPresent(i) ? (
-                <div className="text-red1 font-semibold">{t('dateError')}</div>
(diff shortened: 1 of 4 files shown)


cjhaddad (changes requested):
I pulled this down to test and found some bugs. To repro: select different options from the month filter dropdown. if you select before it has loaded, bugs start to come. you will see an infiinite loop of requests begin and sometimes you see the text in the date filter change and sometimes be in the MM-DD-YYYY format.

katelovescode:
There's some room for refactoring here but it's outside of scope for this ticket.

PR #2057, opened 2021-12-20
Pie 2035/api add update endpoint to attendance

### LINKED ISSUE #2035: I can edit an attendance via the API
## 📋 Summary

We need to add an update endpoint for individual attendances in the `attendances_controller`

AC:
- [ ] I can submit a new check_in for an attendance
- [ ] I can submit a new check_out for an attendance
- [ ] I can change an attendance to an absence
  - [x] #2066
  - [ ] If I change an attendance to an absence, its time_in_care is recalculated based on the schedule for that day
  - [ ] I cannot change an attendance to an absence if the child does not have an active schedule for that weekday
- [ ] time_in_care is recalculated when check_in, check_out, or absence changes (probably already working, just verify)
- [ ] if I change the check_in time to a time that belongs to a different service_day (i.e. before midnight to the day before or after midnight to the day after), the ServiceDay for the new date is created if it doesn't exist already, or the attendance is associated to the existing ServiceDay for that date
  - [ ] All absence requirements above apply here as well; if I change the check_in time to another date, and it's an absence, the above actions happen as well
- [ ] All dates/times are applied *in my timezone* but are saved in UTC in the db
- [ ] Tests are written for all of this behavior

- [X] ALL
- [ ] Nebraska
- [ ] Illinois

## 💻 Implementation Details

Most of this behavior is already accounted for, we just want to write tests to cover edits as well as creation.

twinsfan421:
Closes #2035  It adds an update endpoint to the attendance api controller.  It does not include all the tests in the issue.  Would suggest splitting into separate issue for those tests to be written for coverage on the attendance model.

## 🍂 Before You Submit

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies

### DIFF
--- a/app/models/attendance.rb
+++ b/app/models/attendance.rb
@@ -5,7 +5,7 @@ class Attendance < UuidApplicationRecord
   before_validation :round_check_in, :round_check_out
   before_validation :calc_time_in_care, if: :child_approval
   before_validation :find_or_create_service_day, if: :check_in
-  before_save :remove_absences
+  before_create :remove_absences
 
   belongs_to :child_approval
   belongs_to :service_day
--- a/app/controllers/api/v1/attendances_controller.rb
+++ b/app/controllers/api/v1/attendances_controller.rb
@@ -4,6 +4,8 @@ module Api
   module V1
     # API for user attendances
     class AttendancesController < Api::V1::ApiController
+      before_action :set_attendance, only: %i[update]
+
       # GET /attendances
       def index
         @attendances = policy_scope(Attendance).for_week(filter_date)
@@ -11,6 +13,14 @@ module Api
         render json: AttendanceBlueprint.render(@attendances.includes({ child_approval: :child }))
       end
 
+      def update
+        if @attendance.update(attendance_params)
+          render json: AttendanceBlueprint.render(@attendance)
+        else
+          render json: @attendance.errors, status: :unprocessable_entity
+        end
+      end
+      
       private
 
       def filter_date
@@ -19,6 +29,14 @@ module Api
         else
           Time.current.at_end_of_day
         end
+      end
+
+      def set_attendance
+        @attendance = policy_scope(Attendance).find(params[:id])
+      end
+
+      def attendance_params
+        params.require(:attendance).permit(%i[absence check_in check_out])
       end
     end
   end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -28,7 +28,7 @@ Rails.application.routes.draw do
       get 'profile', to: 'users#show'
       resources :businesses
       resources :children
-      resources :attendances, only: :index
+      resources :attendances, only: %i[index update]
       resources :service_days, only: :index
       resources :attendance_batches, only: :create
       get 'case_list_for_dashboard', to: 'users#case_list_for_dashboard'
--- a/spec/requests/api/v1/attendances_spec.rb
+++ b/spec/requests/api/v1/attendances_spec.rb
@@ -98,4 +98,74 @@ RSpec.describe 'Api::V1::Attendances', type: :request do
       end
     end
   end
+
+  describe 'PUT /api/v1/attendance/:id' do
+    include_context 'with correct api version header'
+
+    let(:attendance) { past_attendances.first }
+    let(:new_check_in) { attendance.check_in + 1.hour }
+    let(:new_check_out) { attendance.check_in + 6.hours }
+
+    context 'when logged in as a non-admin user' do
+      before do
+        sign_in logged_in_user
+      end
+
+      it 'can submit a new check_in for an attendance' do
+        check_in_params = { attendance: { check_in: new_check_in.to_s } }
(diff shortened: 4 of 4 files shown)


katelovescode:
Engineering Check:

AC:
- [ ] I can submit a new check_in for an attendance
- [ ] I can submit a new check_out for an attendance
- [ ] I can change an attendance to an absence
  - [ ] If I change an attendance to an absence, all other attendances for that ServiceDay are *deleted* - we can only have one absence per day, and no other types of attendance if there's an absence
  - [ ] If I change an attendance to an absence, its time_in_care is recalculated based on the schedule for that day
  - [ ] I cannot change an attendance to an absence if the child does not have an active schedule for that weekday
- [ ] time_in_care is recalculated when check_in, check_out, or absence changes (probably already working, just verify)
- [ ] if I change the check_in time to a time that belongs to a different service_day (i.e. before midnight to the day before or after midnight to the day after), the ServiceDay for the new date is created if it doesn't exist already, or the attendance is associated to the existing ServiceDay for that date
  - [ ] All absence requirements above apply here as well; if I change the check_in time to another date, and it's an absence, the above actions happen as well
- [ ] All dates/times are applied *in my timezone* but are saved in UTC in the db
- [ ] Tests are written for all of this behavior

katelovescode:
If you send an absence "null" after you've sent an absence reason, the record deletes itself. 😭 

error_with_submission

katelovescode:
I know those tests belong on the model but I think they should be in scope on this ticket, because the edit behavior is impacted by them.  Let's leave this open til they're written and the behavior is adjusted per the GIF I left above!

katelovescode (changes requested):
Changes requested - behavior + model tests

twinsfan421:
> 

Ok, sounds good. I will keep working on it for all the tests and refactoring for this.

twinsfan421:
> I know those tests belong on the model but I think they should be in scope on this ticket, because the edit behavior is impacted by them. Let's leave this open til they're written and the behavior is adjusted per the GIF I left above!

Ok, sounds good.  I can keep working on it.

app/models/attendance.rb
```diff
@@ -5,7 +5,7 @@ class Attendance < UuidApplicationRecord
   before_validation :round_check_in, :round_check_out
   before_validation :calc_time_in_care, if: :child_approval
   before_validation :find_or_create_service_day, if: :check_in
-  before_save :remove_absences
+  before_create :remove_absences
```
katelovescode, line 8: Let's think about this use case - the likelihood of this might be 0, so feel free to correct me here.
a) is it ever possible in this code to have a service day that has TWO attendances that are both absences?
  - if not, then this is fine, I think
  - if so, we would still want to remove the *other* absence if one of the attendances get edited to be a real attendance vs. an absence, and this trigger wouldn't fire because of the edit

If it's the second case, I think what we want to do instead of changing the hook is change the behavior to remove all absences OTHER than itself when an attendance gets edited from an absence to a real attendance.

Let me know what you think about this, like I said, it's possible we have other logic preventing multiple absences from being added to the same service day, but I don't remember off the top of my head.

twinsfan421: That was a quick temp fix for the gif problem but I will be changing the behavior. There is nothing in the code that prevents a service day of having two attendance absences.  I was also planing on writing a hook to remove not only the other attendances with absences but all other attendances of the service day to satisfy: 

> If I change an attendance to an absence, all other attendances for that ServiceDay are deleted - we can only have one absence per day, and no other types of attendance if there's an absence

I would then change that remove absences hook to a custom validation that prevents two absences for a service day.

This case will require refactoring as well:

> if I change the check_in time to a time that belongs to a different service_day (i.e. before midnight to the day before or after midnight to the day after), the ServiceDay for the new date is created if it doesn't exist already, or the attendance is associated to the existing ServiceDay for that date
>  All absence requirements above apply here as well; if I change the check_in time to another date, and it's an absence, the above actions happen as well

katelovescode: @twinsfan421 one adjustment based on the refactors I just did to unblock the deployment to prod:
```
 If I change an attendance to an absence, its time_in_care is recalculated based on the schedule for that day`
```
This is no longer an AC - instead I put the logic of determining what schedule to apply to a service day *in* the service day; attendances will either calculate their check_out - check_in, or return 0.seconds.  This feels like more fidelity to me on the part of the attendance knowing about itself.  Let me know what you think.

As for your above comments, I'm okay with putting that work in another ticket so we can get the edit endpoint out.  Do you want to write it or do you want me to?

twinsfan421: Ok sounds good, Ya I definitely think returning check_out - check_in, or return 0.seconds for time_in_care is the right call on that.  If you have time to write it, that would be great.  Otherwise I can do it when I get back.  I am I about to head to the airport and fly home for the holidays.  I will be available after the new year to pick back up on it.  Thanks!

katelovescode: https://github.com/pieforproviders/pieforproviders/issues/2080 Made the ticket here, just used your comment since it was clearer than rewriting.  :)  Thanks!

PR #2115, opened 2022-01-08
Pie 1312/multiple attendance entry

cjhaddad:
This PR closes 1312 -- adding the frontend functionality to have two sets of attendance inputs in each cell on the table on the Attendance edit page. 

## 🍂 Before You Submit

* [X] Does your PR contain any required translations?

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies

## 🧳 Steps to test
When on the attendance edit page, ensure that you can select "add check-in time" and another set of attendance inputs appears. Also, make sure that when you either select remove or "absent" the second set of inputs is removed and only one attendance object for that cell is sent to the backend. Generally, ensure that all functionality on this edit page is working and the data is being sent to the backend correctly and as expected given what is shown on the frontend. 

Note: this required a shift in how the frontend's data structure of this table, make sure to regression test to ensure this change did not break anything.


client/src/Attendance/Attendance.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-debugger */
```
katelovescode, line 1:
```suggestion
```

katelovescode (changes requested):
When I try this locally, only the first attendance entered is saved; the second box doesn't persist to the db.

Also, can we write a test somewhere that tests the expectation of what we're sending to the backend?  i.e. I expect when I fill out x input field, we're hitting the api with y data?  I don't know if that's a front-end best practice test or not, so feel free to ignore me if that's not really how it should be tested :D

katelovescode (approved):
Works great, thank you!

Issue #2045, opened 2021-12-16
Speed up our expensive queries - unique indexes, etc.


katelovescode:
## 📋 Summary

Examples:
- WS ID on Attendances should be Unique IDs
- Do we need to change to autoincrementing IDs rather than UUIDs for speed.  Maybe for attendances?  Or Service Days, or just the larger tables?

Top Ten Min Query Length (between 5ms and 0.05ms):
```
SELECT "attendances".* FROM "attendances" INNER JOIN "service_days" ON "attendances"."service_day_id" = "service_days"."id" WHERE "attendances"."absence" IS NOT NULL AND ("service_days"."child_id" = $1 AND (check_in BETWEEN $3 AND $4) OR "attendances"."service_day_id" = $2)
SELECT "attendances".* FROM "attendances" INNER JOIN "child_approvals" ON "child_approvals"."id" = "attendances"."child_approval_id" WHERE "attendances"."absence" IS NOT NULL AND ("child_approvals"."child_id" = $1 AND (check_in BETWEEN $3 AND $4) OR "attendances"."service_day_id" = $2)
SELECT "service_days".* FROM "service_days" INNER JOIN "attendances" ON "attendances"."service_day_id" = "service_days"."id" INNER JOIN "children" ON "children"."id" = "service_days"."child_id" INNER JOIN "businesses" ON "businesses"."id" = "children"."business_id" WHERE "service_days"."id" IN ($1, $2, $3, $4, $5) AND "businesses"."state" = $6 GROUP BY "service_days"."id" HAVING (sum("attendances"."total_time_in_care") between $7 and $8 OR sum("attendances"."total_time_in_care") between $9 and $10 OR sum("attendances"."total_time_in_care") > $11)
SELECT "service_days".* FROM "service_days" INNER JOIN "attendances" ON "attendances"."service_day_id" = "service_days"."id" INNER JOIN "children" ON "children"."id" = "service_days"."child_id" INNER JOIN "businesses" ON "businesses"."id" = "children"."business_id" WHERE "service_days"."id" IN ($1, $2, $3, $4) AND "businesses"."state" = $5 GROUP BY "service_days"."id" HAVING (sum("attendances"."total_time_in_care") between $6 and $7 OR sum("attendances"."total_time_in_care") between $8 and $9 OR sum("attendances"."total_time_in_care") > $10)
[...]

PR #2159, opened 2022-01-20
[PIE-2045] [Partial] Storing some data on the ServiceDay for better performance

katelovescode:

This is some work towards #2045 - I'm storing the schedule and the total_time_in_care directly on the ServiceDay rather than dynamically calculating it at runtime on the dashboard.  This code adds an after_save_commit hook to attendances that calls a TotalTimeInCareCalculator service object that updates the Service Day at the same time, segmented by state (i.e. we have a specific one for Nebraska) in hopes of modeling this pattern in the future.  Using a service object to update records besides the model that's acting on it is also a pattern I'd like to repeat.

Changes to the code in this PR:
- Adds profiling gems and a `profile` environment to run `ruby-prof` reports.  To use this, you'll need to:
  - set up a `secret_key_base` in yaml by editing credentials for the profile environment only: `RAILS_ENV=profile be rails credentials:edit`
    - The file needs only one line: `secret_key_base: test` for example
  - `RAILS_ENV=profile SECRET_KEY_BASE=test be rails s` for API only or `RAILS_ENV=profile SECRET_KEY_BASE=test be rails start` to spin up the front end (though I haven't tested with the front-end)
  - Hit the API with Postman or the front end and the gem will generate profiling reports in `tmp/profile/` - note if you hit the same endpoint twice the results will be overwritten, not appended
  > NB: This looks neat: https://github.com/tleish/ruby-prof-rails
- Passes the service days into the Nebraska::DashboardCase.new call in the Child Blueprint, which cut performance by 87% locally
- Stores the `total_time_in_care` and `earned_revenue` on `ServiceDay` to avoid dynamic calculation - this might not have helped much 👿 
- Associates a ServiceDay with a schedule directly to be able to pull scheduled duration faster - this might not have helped much 👿 
- Set up Attendance and Schedule changes to re-run total_time_in_care and earned_revenue calculations, and Schedule changes ensure the ServiceDays are all associated correctly, all in background jobs - this seems like a BEAST of a way to handle this behavior, so I'm sure this will be ripe for refactoring soon - this might not have helped much 👿 
- Stops running every ServiceDay object through CalculatedServiceDay and instead only uses it for projected future ServiceDays instead of existing attendances and absences - this might not have helped much 👿 
- Fixed the eager-loading that happens w/ the User to not apply when only pulling a *single* user vs. an admin based on Bullet output in the users_controller request spec
- Changes all money fields to use the Money library (either via field definition or type casting)
- @abugosh found a ton of places to reduce multiple calls to time/schedules
- @abugosh found several places to do hard `.find` queries on results instead of queries that made duplicate DB calls
- Added LOTS of `perform_enqueued_jobs` and service day reloads throughout the specs
- Broke the dashboard case blueprint spec into MUCH smaller expectation groups with consistent shared_contexts that built on how I had it before but makes it tons easier to diagnose problems or errors.

We still have lots of work to do but this is a good start.

## 🍂 Before You Submit

## 🛷 Deployment

* [X] Migrations

## 🧵 Steps to set up locally

```
bundle exec rails db:migrate
```

## 🧳 Steps to test

The dashboard should work the same, and if my theory is correct, some % faster.


katelovescode:
@abugosh following up - I moved schedule update recalculation to its own service that calls the total time in care calculator.  Also I refactored that enormous spec using shared examples; it gets rid of flakiness and weirdly-timed callbacks or having to reload multiple times through the same spec which seems to have unexpected behavior.  I do have one spec at the end with schedule changes that cause the recalculation to happen but those are working as expected so I'm assuming there are just not enough callback loops to cause order confusion now that I've refactored more out into a service.  I'm taking a break now, but in this PR I still want to handle calculating earned revenue and storing it on the ServiceDay as another service object, so I'll come back and do that in a few hours.

abugosh:
Based on the profiler, currently we are spending the most time in time calculations. If we can simplify 

```ruby
schedule.weekday == weekday &&
schedule.effective_on <= date &&
(schedule.expires_on.nil? || schedule.expires_on > date)
```

in all of the places we call it to stop doing expensive time/timezone operations, it'll take another chunk out of time spent.

app/blueprints/business_blueprint.rb
```diff
@@ -15,6 +15,19 @@ class BusinessBlueprint < Blueprinter::Base
   view :nebraska_dashboard do
     field :name
     exclude :id
+    # TODO: Multithreading? Pseudocode
```
twinsfan421, line 18: I wouldn't recommend multithreading for a smaller ruby on rails application

PR #2162, opened 2022-01-21
[PIE-2121] Fix multiple api calls in dashboard

### LINKED ISSUE #2121: Dashboard makes API call twice
## 👉🏼 Steps to Reproduce

Explanation in the following slack conversation:
Screen Shot 2022-01-10 at 10 23 55 AM

## 🇺🇸 States this bug is happening in
- [X] ALL
- [ ] Nebraska
- [ ] Illinois

## 🧥Expected behavior

The API will only be hit one time on `case_list_for_dashboard` - it's an expensive query and shouldn't be called unless strictly necessary

## 💻 Browser Info

Please complete the following information if accessing the app from a computer:

- Operating System: [e.g. iOS]
- Browser [e.g. Chrome, Safari]

anniekao:
Fixes #2121

## 🍂 Before You Submit

## 🧳 Steps to test
- Open developer tools and go to network tab
- Load dashboard, which should then call case_list_for_dashboard once
- Choose a different month from the dropdown list and the endpoint is called again with the selected month as the query param

## 🕯 Caveats, concerns
The test for this fix hasn't been written. I wanted to get feedback on my approach before going forward since I'm not 100% sure about it.
The plan was to implement something like this in the tests for DashboardTitle:
```
it('makes a call to get dashboard data by selected month', () => {
    // Check for text in the dropdown, e.g. Mar. 2022
    // Click on the dropdown
    // Select a different month
    // Check for text, e.g. Feb. 2022 
    // Check if /case_list_for_dashboard has been called
  })
```


cjhaddad:
@anniekao This looks good to me and also your testing approach also sounds good. You could also be explicit in your test to test that the  /case_list_for_dashboard endpoint has only been called once.

cjhaddad (changes requested):
Aside from updating the PR to include the test you mentioned adding, this looks good to me. Once that test is added, we can merge this.

Thank you!!

katelovescode:
@anniekao Just wanted to check on this, do you think you'll have time to add the test soon?  Thanks!

anniekao:
Hey @katelovescode -- I've started work on the test, but it's taking me longer than expected because of issues I'm having interacting with Ant Design's Select component. I've worked out how to get the drop down to appear and target an option, but for some reason the mock call isn't being detected and the test is failing. I'm going to have another look at it on the weekend. I'll post an update on Monday.

Just as a reference here are some threads that talk about the issue with the Select in Ant Design v4 and some of the solutions that have worked for some:
- [https://github.com/ant-design/ant-design/issues/23009](https://github.com/ant-design/ant-design/issues/23009)
- [https://stackoverflow.com/questions/61080116/ant-design-v4-breaks-react-testing-library-tests-for-select-and-autocomplete](https://stackoverflow.com/questions/61080116/ant-design-v4-breaks-react-testing-library-tests-for-select-and-autocomplete)
- [https://github.com/ant-design/ant-design/issues/22074](https://github.com/ant-design/ant-design/issues/22074)

katelovescode:
@anniekao Sounds good - looking forward to an update!  Also, I'd timebox this at another hour at most; if you can't write the test we have the comments here about why and I think we can approve and move forward for now.  Thank you!

anniekao:
@katelovescode, @cjhaddad the test is working and calls `getDashboardData` after the month filter changes, just missing a minor piece at the end checking the UI that the dropdown text content/month has updated. I think would require a rerender of the component with updated dates, but I ran out of time.  Let me know if there's any thing missing or just comments in general!

katelovescode:
@anniekao if you think you can get it done in a short while, go ahead and finish!  Otherwise let me know and we'll merge as-is!

anniekao:
@katelovescode thanks for waiting! I was able to get the rerender to work, so thing should be good to go.

PR #2255, opened 2022-02-16
fixed roudning bug and set max for payment input

davidcandeau:
## 🍂 Before You Submit

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies

### DIFF
--- a/client/src/Dashboard/DashboardTitle.js
+++ b/client/src/Dashboard/DashboardTitle.js
@@ -237,7 +237,7 @@ export default function DashboardTitle({
     >
       <p>
         {t('paymentSuccessText')} {t(previousMonth)} {previousMonthYear}{' '}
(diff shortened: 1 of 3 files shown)


client/src/Dashboard/DashboardTitle.js
```diff
@@ -164,7 +164,7 @@ export default function DashboardTitle({ dates, setDates, makeMonth }) {
             className="record-payment-button"
             onClick={addPayment}
           >
-            {t('recordPaymentOf')} ${totalPayment.toFixed()}
+            {t('recordPaymentOf')} ${totalPayment}
```
davidcandeau, line 167: Will all the calculations initially be up the 2nd decimal point? Otherwise, we would need to round her for the initial load?

davidcandeau: @katelovescode

katelovescode (approved):

Issue #2212, opened 2022-02-03
Tweak fields on sign up page 


csprayregen:
## 🔬 Experiment Details

If we change our multi-step signup process form from two websites into one, we will increase visits to our and decrease bounce rates from the signup page.

**Out of scope**
- "Create an account" header
- Login button prompt on the bottom of the page

## 🇺🇸 States this change applies to
- [X] ALL
- [ ] Nebraska
- [ ] Illinois

## 🧶 Acceptance Criteria

**Changes to fields**
- [ ] Make cell number mandatory
- [ ] Add State field (dropdown)
- [ ] Remove "password confirmation" field
- [ ] Add mandatory multi-step question "Help us understand your current experience managing the subsidy" (use[ radio button "radio style" component](https://ant.design/components/radio/#components-radio-demo-radiobutton) from Ant)
- [ ] Add non-mandatory question " What are you hoping to get from using Pie for Providers"? 
- [ ] This works in English and Spanish 

**UI & Copy Changes**
- [ ] Change width of & Sign Up CTA button to green
- [ ] Add text below sign up button: 
Once you create your account, your dashboard will remain blank until we update it with more information.
To learn more about Pie for Providers, go to: www.pieforproviders.com
- [ ] Add H1 that says "Welcome to Pie for Providers!" 

**Nice to haves**
- [ ] Change selected state style of Preferred Language
- [ ] Add radius to radio buttons so they can be consistent w the other input fields

## 🥧 Implementation Details
[Link to Figma](https://www.figma.com/file/GAYeTfOa7umyAhssCHUf8d/P4P-2.0?node-id=50579%3A57543) 

Spanish translations for new language: https://docs.google.com/spreadsheets/d/1d0RPc9PXvTCFpq8bfkZxgNL35MEpX5s8EvBwqex8564/edit#gid=590755784

rebeccakarasiktw:
FYI @cjhaddad updated ticket w what we discussed today

rebeccakarasiktw:
Hey @cjhaddad I found some small things I've added the comments in the screenshots here. LMK if you want to discuss
<img width="1440" alt="Screen Shot 2022-03-23 at 20 20 46" src="https://user-images.githubusercontent.com/26929739/159824094-961fada9-63a7-41f0-b261-06514ba9a37e.png">
!
[...]

PR #2264, opened 2022-02-18
[PIE-2212] Signup form changes - API

katelovescode:

Adds `state` to the user model, adds survey questions to the user model.  Backend work for #2212 

## 🍂 Before You Submit

## 🛷 Deployment

* [X] Migrations

## 🧵 Steps to set up locally

```
bundle exec rails db:migrate
```

## 🧳 Steps to test

- Send a registration request to the API using the state param.  It should successfully create a user.
- Send a registration request to the API without the state param.  It should return an error.
- Send a registration request to the API with the state param and survey question answers (one of "True", "Mostly True", "Mostly False", "False").  It should successfully create a user.
- Send a registration request to the API with the state param and a disallowed value for the survey question answers (anything but "True", "Mostly True", "Mostly False", or "False").  It should return an error.


katelovescode:
@cjhaddad should be fixed!

Issue #1170, opened 2021-04-14
I can edit attendance records for the kids in my care


csprayregen:
## ✨ What new feature do you want to see?
I can make changes to the attendance records I have added for the kids in my care (see #1125 for adding records and #1169 for viewing records).

## 🧶 Acceptance Criteria

- [x] I can edit the check in time
- [x] I can edit the check out time
- [x] I can change a status to "absent"
- [x] I can change the reason for an absence
- [x] If a record came from WS, I cannot edit it. I am directed to edit it in WS --> pending discussion of WS attendance in #1169 
- [x] - [ ] I can add a second check in/out time
- [x] If there are multiple check in/out times, I can edit both 
-- talking with @cjhaddad we decided to add these ACs: --
- [ ] ~If I put a check out time that is earlier than my check in time, I will see a message that stops me from doing so, like this: ~
![image](https://user-images.githubusercontent.com/26929739/152246673-ee205086-5422-43e3-9af1-482c8058f9b2.png) --> CS: per 2/7 team meeting @katelovescode to create separate ticket for this
 - [ ] If I haven't entered a check in & check out time OR if I haven't checked Absent, the "Save" button will be disabled --> CS: per 2/7 team meeting, button will be disabled unless I have a check in time OR checked Absences. I do not need a check out time for this validation

## 🖍️ Implementation Details

[**figma link**](https://www.figma.com/file/uCveVQQnYcw3MsenD1PNcG/P4P-2.0?node-id=5556%3A32631)

> **NOTE TO ENGINEERS**
> 
> Please ensure that you are running Google Lighthouse and/or WebAIM (Wave) on any UI elements in your pull request to check for accessibility issues.
> 
> If it is out of scope to address those accessibility issues in your pull request (because the issue is in a parent component and needs refactoring to work, for example), **you are responsible for noting those issues in your pull request and writing bug tickets to address the issues - make sure you link the new tickets in your pull request**

katelovescode:
Blocked by #1312

rebeccakarasiktw:
Per conversation with @cjhaddad, adding to scope of this card:
* Update to AC so that the buttons say "Save" and "Cancel"
* "Save" button should be disabled until both check in / check out are present OR you choose absent
* Add error message w/in modal if check-out time is earlier than check-in time.

TO THINK ABOUT MORE 
* How do we best communicate to the user once the updates have been made and saved. Some kind of message to confirm your save went thru?

[...]

PR #2269, opened 2022-02-23
Removes empty service days from response to API

katelovescode:

Supports #1170 - removes empty service days from API endpoint

## 🍂 Before You Submit

## 🧳 Steps to test

Delete all attendances for a service day.  It should no longer show up in the endpoint returning service days.  It should be deleted.

### DIFF
--- a/app/controllers/api/v1/service_days_controller.rb
+++ b/app/controllers/api/v1/service_days_controller.rb
@@ -7,7 +7,7 @@ module Api
       # GET /service_days
       def index
         @service_days = policy_scope(
-          ServiceDay.includes(child: { business: :user })
+          ServiceDay.joins(:attendances).includes(child: { business: :user })
           .includes(attendances: { child_approval: :child })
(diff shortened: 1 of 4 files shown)


twinsfan421 (commented):
@katelovescode The changes look good for the destroy_empty_service_day callback.  Did you mean to remove the remove_old_service_day destroy callback?   Wouldn't you still have the problem of an empty service day if the check_in time is edited to a different service day and it leaves the service day with no attendances?

katelovescode:
> @katelovescode The changes look good for the destroy_empty_service_day callback. Did you mean to remove the remove_old_service_day destroy callback? Wouldn't you still have the problem of an empty service day if the check_in time is edited to a different service day and it leaves the service day with no attendances?

Ah, crap, good point, I'll add that back in.  My brain was like, "Oh, the service day will be deleted" but it won't because the attendance isn't being destroyed, it's being edited :/ WOMP.  Thanks for catching this!

PR #2279, opened 2022-02-23
Adds back in the logic to remove old service days from attendances

katelovescode:
Adds back in the logic to remove old service days from attendances that have been moved, when it's the last attendance on that day


katelovescode:
cc @twinsfan421

Issue #2128, opened 2022-01-12
On edit attendance: call to children endpoint instead of case_list_for_dashboard


katelovescode:
## 📋 Summary

Currently when you visit the Add/Edit Attendance view, the frontend makes a call to case_list_for_dashboard to get information about the children in the user's care; instead, we should get that info from the /children endpoint because the /case_list_for_dashboard endpoint is an expensive query.

## 🇺🇸 States this change applies to
- [X] ALL
- [ ] Nebraska
- [ ] Illinois

## 💻 Implementation Details

If anything is missing from the /children endpoint that's needed for this view, please write a backend ticket to add info to the API.

PR #2361, opened 2022-03-15
Changes the Edit Attendance Page to call to the Children endpoint

katelovescode:
Changes the Edit Attendance Page to call to the Children endpoint, which does not limit the list to children with active approvals.

No ticket, this fixes a request from a customer directly.  However, this covers a prior ticket: #2128 

## 🍂 Before You Submit

## 🧳 Steps to test

Ensure you have one student with no active approvals.  The student should show up on Edit Attendance, but not on the Dashboard

## 🕯 Caveats, concerns

@cjhaddad the styling looked off to me when I ran it locally but my computer has been slow and buggy today, so it might have been a side effect.  Can you double-check?

### DIFF
--- a/app/blueprints/business_blueprint.rb
+++ b/app/blueprints/business_blueprint.rb
@@ -3,9 +3,9 @@
 # Serializer for businesses
 class BusinessBlueprint < Blueprinter::Base
   identifier :id
+  field :name
 
   view :illinois_dashboard do
-    field :name
     exclude :id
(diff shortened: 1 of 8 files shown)


katelovescode:
> code and styling in the browser-wise this looks good to me! how hough can I confirm that the "see if a student with no approvals doesn't show in dashbaord, but shows up in edit attendance" scenario works?

go into the console and do something like this:
```
child = Child.first
child.approvals.count == 1 # make sure the kid has only one approval - should be true
child.approvals.first.update!(expires_on: Time.current - 1.day)
child.reload
```
then reload the page

Issue #2374, opened 2022-03-15
Change backend to return child records/case records sorted by last name


katelovescode:
## 📋 Summary

Case records should be returned from the /children endpoint sorted alphabetically by last name by default.

AC:
[ ] The Add Attendance table sorts alphabetically, by Child Last Name
[ ] This works in English and Spanish

## 🇺🇸 States this change applies to
- [X] ALL
- [ ] Nebraska
- [ ] Illinois

PR #2428, opened 2022-04-01
change backend to return children sorted by last name

twinsfan421:
 #2374  This would update the dashboard db calls to order the children by last_name

## 🍂 Before You Submit

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies

### DIFF
--- a/app/controllers/api/v1/children_controller.rb
+++ b/app/controllers/api/v1/children_controller.rb
@@ -11,9 +11,9 @@ module Api
       def index
         # binding.pry_remote
         @children = if params[:child].present? && child_params[:site]
-                      policy_scope(Child.includes(business: :user).where(business: Business.find(child_params[:site])))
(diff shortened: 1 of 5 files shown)


katelovescode:
This looks good - per your question in Slack, I do think it'd be great if you could write tests as well.

twinsfan421:
> This looks good - per your question in Slack, I do think it'd be great if you could write tests as well.

I pushed up some tests and fixes.

app/controllers/api/v1/children_controller.rb
```diff
@@ -58,6 +51,16 @@ def set_child
         @child = policy_scope(Child).find(params[:id])
       end
 
+      def set_children
+        # binding.pry_remote
```
katelovescode, line 55:
```suggestion
```

katelovescode (approved):

PR #2432, opened 2022-04-04
Dcandeau/1622/no edit message

davidcandeau:
## 🍂 Before You Submit

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies


katelovescode (changes requested):
My initial instinct is that the trigger to check if a payment has been recorded for this month yet belongs in the paymentModal component itself.  Given that eventually we'll have a drop down to select prior months, I'm not sure if there's a practical way to store "isPaymentSubmitted" from the response of submitting the payment; you might just need to make the request every time the modal opens.  

As implemented now, the page requires a refresh to display the "already submitted" message - i.e. if I submit a payment, click okay on the success message, then click "record new > payment" again, I can submit more payments.

@cjhaddad what are your thoughts on this?

cjhaddad:
> My initial instinct is that the trigger to check if a payment has been recorded for this month yet belongs in the paymentModal component itself. Given that eventually we'll have a drop down to select prior months, I'm not sure if there's a practical way to store "isPaymentSubmitted" from the response of submitting the payment; you might just need to make the request every time the modal opens.
> 
> As implemented now, the page requires a refresh to display the "already submitted" message - i.e. if I submit a payment, click okay on the success message, then click "record new > payment" again, I can submit more payments.
> 
> @cjhaddad what are your thoughts on this?



> My initial instinct is that the trigger to check if a payment has been recorded for this month yet belongs in the paymentModal component itself. Given that eventually we'll have a drop down to select prior months, I'm not sure if there's a practical way to store "isPaymentSubmitted" from the response of submitting the payment; you might just need to make the request every time the modal opens.
> 
> As implemented now, the page requires a refresh to display the "already submitted" message - i.e. if I submit a payment, click okay on the success message, then click "record new > payment" again, I can submit more payments.
> 
> @cjhaddad what are your thoughts on this?

Yes, it looks like it definitely makes sense to put the call to check if a payment is submitted needs to occur each time the Modal opens. The behavior we see now is because the call happens once and is stored in the state but not updated again after the payment modal closes and has successful submission.

cjhaddad:
> My initial instinct is that the trigger to check if a payment has been recorded for this month yet belongs in the paymentModal component itself. Given that eventually we'll have a drop down to select prior months, I'm not sure if there's a practical way to store "isPaymentSubmitted" from the response of submitting the payment; you might just need to make the request every time the modal opens.
> 
> As implemented now, the page requires a refresh to display the "already submitted" message - i.e. if I submit a payment, click okay on the success message, then click "record new > payment" again, I can submit more payments.
> 
> @cjhaddad what are your thoughts on this?



> My initial instinct is that the trigger to check if a payment has been recorded for this month yet belongs in the paymentModal component itself. Given that eventually we'll have a drop down to select prior months, I'm not sure if there's a practical way to store "isPaymentSubmitted" from the response of submitting the payment; you might just need to make the request every time the modal opens.
> 
> As implemented now, the page requires a refresh to display the "already submitted" message - i.e. if I submit a payment, click okay on the success message, then click "record new > payment" again, I can submit more payments.
> 
> @cjhaddad what are your thoughts on this?

Yes, it looks like it definitely makes sense to put the call to check if a payment is submitted needs to occur each time the Modal opens. The behavior we see now is because the call happens once and is stored in the state but not updated again after the payment modal closes and has successful submission.

client/src/Payment/paymentModal.js
```diff
@@ -1,19 +1,23 @@
 import React, { useEffect, useState } from 'react'
```
katelovescode, line 1: @davidcandeau small note for consistency, this file should be named `PaymentModal` - not sure how that slipped past us.

app/models/attendance.rb
```diff
@@ -132,16 +132,16 @@ def delete_or_mark_absent
...
-#  created_at        :datetime         not null
-#  updated_at        :datetime         not null
-#  child_approval_id :uuid             not null
-#  service_day_id    :uuid
-#  wonderschool_id   :string
+#  id                                                       :uuid             not null, primary key
+#  check_in                                                 :datetime         not null
+#  check_out                                                :datetime
+#  deleted_at                                               :date
+#  time_in_care(Calculated: check_out time - check_in time) :interval         not null
```
katelovescode, line 139: Don't commit these changes; I'm working to get annotate to stop doing it

app/models/nebraska_dashboard_case.rb
```diff
@@ -27,7 +27,7 @@ class NebraskaDashboardCase < UuidApplicationRecord
 #  full_days_remaining        :integer          default(0), not null
 #  hours                      :float            default(0.0), not null
 #  hours_remaining            :float            default(0.0), not null
-#  month                      :datetime         default(Tue, 10 May 2022 19:40:48.447958000 UTC +00:00), not null
```
katelovescode, line 30: Ditto here on unrelated changes

client/src/Payment/paymentModal.js
katelovescode: :boop: on this one, can you rename this file, please?

spec/factories/attendances.rb
```diff
@@ -102,16 +102,16 @@
...
-#  created_at        :datetime         not null
-#  updated_at        :datetime         not null
-#  child_approval_id :uuid             not null
-#  service_day_id    :uuid
-#  wonderschool_id   :string
+#  id                                                       :uuid             not null, primary key
+#  check_in                                                 :datetime         not null
+#  check_out                                                :datetime
+#  deleted_at                                               :date
+#  time_in_care(Calculated: check_out time - check_in time) :interval         not null
```
katelovescode, line 109: Ditto here, uncommit pls

db/schema.rb
```diff
@@ -191,7 +191,7 @@
   end
 
   create_table "nebraska_dashboard_cases", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
-    t.datetime "month", default: "2022-05-10 19:40:48", null: false
+    t.datetime "month", default: "2022-05-11 19:17:16", null: false
```
katelovescode, line 194: None of this should be committed, also same with pie_erd (I can't comment on that file because it's binary)

katelovescode (changes requested):
Please uncommit the files as requested, but otherwise this works great.  I'll approve once the extraneous files are pulled out.

katelovescode:
Reopening as a separate branch.

Issue #2431, opened 2022-04-01
View Attendance Tags - show billable numbers


katelovescode:
## 📋 Summary

Update the tags on the view attendance page to show the correct (rounded) billing number of hours and days so providers can use that page to directly bill.

- Wants to see 2.75 hours, not 2 hours and 45 minutes. If I’m going to manually add it up, might as well stay in the system I have.

## 🇺🇸 States this change applies to
- [X] ALL
- [ ] Nebraska
- [ ] Illinois

## 💻 Implementation Details

The rounding we use is already in place in the dashboard calculations; we just want to display these numbers in the existing "hours" and "full day" tags.  The service days endpoint returns tags "hourly" and "daily" - we want to return (for example) "4.75 hours" and "1 full day".  See `app/services/nebraska/daily/hours_duration_calculator.rb` and `app/services/nebraska/daily/days_duration_calculator.rb`, but the rule in words is:
- up to 6 hours is charged hourly (rounded up to the nearest 15 minutes, display as decimal)
- between 6 and 10 hours is 1 full day
- above 10 hours is 1 full day plus the additional hours above 10 (rounded up to the nearest 15 minutes, display as decimal)
- the max is 18 hours (1 full day + 8 hours) - anything over that should display as if it were 18 hours

rebeccakarasiktw:
Hi @twinsfan421 @katelovescode this looks good but can we make a couple of small tweaks, see comments / screen shot below:

<img width="312" alt="Screen Shot 2022-04-19 at 13 21 57" src="https://user-images.githubusercontent.com/26929739/164070701-dfc40dd1-41b3-4e57-b568-8cb6cc2cc101.png">

- If the hour is a full hour i.e. 2 or 3 hours, can we get rid of the ".0" afterwards?
- If it is just 1 hour, can it say "1 hour" instead of "1.0 hours" ie make the word hours singular?

PR #2458, opened 2022-04-12
view attendance tags show billable numbers

twinsfan421:
#2431 This would update the tags on the view attendance page to show the calculated amount for both daily and hourly.

## 🍂 Before You Submit

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -11,6 +11,8 @@
 # IgnoredMethods: refine
 Metrics/BlockLength:
   Max: 208
+  Exclude:
+    - 'spec/models/service_day_spec.rb'
 
 # Offense count: 3
 # Configuration parameters: CountComments, Max, CountAsOne, ExcludedMethods, IgnoredMethods.
--- a/app/models/service_day.rb
+++ b/app/models/service_day.rb
@@ -111,20 +111,20 @@ class ServiceDay < UuidApplicationRecord
 
   def tag_daily_amount
     return unless tag_daily
-    
+
     Nebraska::Daily::DaysDurationCalculator.new(total_time_in_care: total_time_in_care).call&.to_f&.to_s
   end
 
   def tag_hourly
     return unless state == 'NE'
 
-    hourly? || daily_plus_hourly? || daily_plus_hourly_max? ? "hourly" : nil
+    hourly? || daily_plus_hourly? || daily_plus_hourly_max? ? 'hourly' : nil
   end
 
   def tag_daily
     return unless state == 'NE'
 
-    daily? || daily_plus_hourly? || daily_plus_hourly_max? ? "daily" : nil
+    daily? || daily_plus_hourly? || daily_plus_hourly_max? ? 'daily' : nil
   end
 
   def tag_absence
--- a/app/models/service_day.rb
+++ b/app/models/service_day.rb
@@ -120,7 +120,7 @@ class ServiceDay < UuidApplicationRecord
   end
 
   def tag_daily_amount
-    Nebraska::Daily::DaysDurationCalculator.new(total_time_in_care: total_time_in_care).call&.to_f&.to_s
+    Nebraska::Daily::DaysDurationCalculator.new(total_time_in_care: total_time_in_care).call&.to_s
   end
 
   def tag_absence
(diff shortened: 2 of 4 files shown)


katelovescode:
Screen Shot 2022-04-12 at 1 57 31 PM
Can we group these tags together as `8.0 hours` and `1.0 days`

twinsfan421:
> Screen Shot 2022-04-12 at 1 57 31 PM
> 
> Can we group these tags together as `8.0 hours` and `1.0 days`

I just sent and update for that

katelovescode:
> > Screen Shot 2022-04-12 at 1 57 31 PM
> > Can we group these tags together as `8.0 hours` and `1.0 days`
> 
> @katelovescode I just sent and update for that but I noticed that since you have ` "daily": "full day", "hourly": "hours",` in the index json, by combining them, it doesn't recognize it and shows up as hourly and daily instead of hours and full day. Would you want me to revert back to how it was and change the front end style or how do you want to handle it?

What I think you're going to want to do is change the backend to return `8.0 hours` and `1.0 full days` (for example) and then split the string and translate it on the front end.  So something like:
```js
/* client/src/AttendanceView/AttendanceView.js:195 */
<div className="flex justify-center">
                {(matchingServiceDay.tags || []).map((tag, i) => (
                  <div
                    key={i}
                    className={`bg-green2 text-green1 box-border p-1 mt-1 ${
                      i > 0 ? 'ml-1' : null
                    }`}
                  >
                    {`${tag.split(' ')[0]} ${t(`${tag.split(' ')[1].toLowerCase()}`)}
                  </div>
                ))}
              </div>```
              
Keeping in mind this is pseudocode concatenating might be a little trickier than that.

Also @rebeccakarasiktw or @cjhaddad - is it okay to change that second tag to `days` instead of `full days`?  Let me know what you think.

twinsfan421:
> > > Screen Shot 2022-04-12 at 1 57 31 PM
> > > Can we group these tags together as `8.0 hours` and `1.0 days`
> > 
> > 
> > @katelovescode I just sent and update for that but I noticed that since you have ` "daily": "full day", "hourly": "hours",` in the index json, by combining them, it doesn't recognize it and shows up as hourly and daily instead of hours and full day. Would you want me to revert back to how it was and change the front end style or how do you want to handle it?
> 
> What I think you're going to want to do is change the backend to return `8.0 hours` and `1.0 full days` (for example) and then split the string and translate it on the front end. So something like:
> 
> ```js
> /* client/src/AttendanceView/AttendanceView.js:195 */
> <div className="flex justify-center">
>                 {(matchingServiceDay.tags || []).map((tag, i) => (
>                   <div
>                     key={i}
>                     className={`bg-green2 text-green1 box-border p-1 mt-1 ${
>                       i > 0 ? 'ml-1' : null
>                     }`}
>                   >
>                     {`${tag.split(' ')[0]} ${t(`${tag.split(' ')[1].toLowerCase()}`)}
>                   </div>
>                 ))}
>               </div>
> ```
> 
> Keeping in mind this is pseudocode, and concatenating might be a little trickier than that.
> 
> Also @rebeccakarasiktw or @cjhaddad - is it okay to change that second tag to `days` instead of `full days`? Let me know what you think.

@katelovescode Ok I just pushed up a fix for that.

rebeccakarasiktw:
Hi @katelovescode I think it would be preferible to keep it as "full day" but can we please take out the ".0" for the "full day" because as I understand it there isn't a way for there to be more than 1, 2, etc full days.

twinsfan421:
> Hi @katelovescode I think it would be preferible to keep it as "full day" but can we please take out the ".0" for the "full day" because as I understand it there isn't a way for there to be more than 1, 2, etc full days.

I just pushed up a commit to remove the ".0" from the full day but can change it back if not preferred.  It would look like this:
Screen Shot 2022-04-13 at 7 43 51 PM

katelovescode:
> > Hi @katelovescode I think it would be preferible to keep it as "full day" but can we please take out the ".0" for the "full day" because as I understand it there isn't a way for there to be more than 1, 2, etc full days.
> 
> I just pushed up a commit to remove the ".0" from the full day but can change it back if not preferred. It would look like this: Screen Shot 2022-04-13 at 7 43 51 PM

The good news is, this will never need to say "2 full days" in NE because they top it at 18 hours :) Looks great, thanks for the speedy turnaround!

katelovescode (approved):

PR #2476, opened 2022-04-19
No ticket/fix scope for user policy

katelovescode:

I don't want to limit the user endpoint to NE for admins; I'll need to limit the user blueprint instead.

## 🍂 Before You Submit

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies


katelovescode:
@nicorinn your PR helped me find a bug/undesired behavior - I'll need to make an additional change on the `main` branch once we're merged in but I think this is important to get into your PR because it makes `/profile` and `/users/:id` work the same way for admin users.

PR #2602, opened 2022-05-13
Pie 1698/loading state

cjhaddad:
## 🍂 Before You Submit

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies


katelovescode:
I'm not seeing the loading bar filling up at the bottom, should that be there?
noloadingbar

cjhaddad:
> I'm not seeing the loading bar filling up at the bottom, should that be there? noloadingbar [ ](https://user-images.githubusercontent.com/8364647/169446207-311c041e-e9da-4a59-8dec-6bfa401bdf66.gif) noloadingbar [ ](https://user-images.githubusercontent.com/8364647/169446207-311c041e-e9da-4a59-8dec-6bfa401bdf66.gif)

yes it should! let me take a look.

cjhaddad:
@katelovescode yes, it should! I will look. my guess is this could be a case where it is a quick load and the isloading state changes before the dom updates the progress bar. 👁️

cjhaddad:
@katelovescode I am having trouble repro-ing. can you make a video? or perhaps we can look at it next week?

katelovescode:
So it looks like the loading bar behavior is inconsistent for me.  Sometimes it's at 0, sometimes it's at 100, but I never see it move.
May-20-2022 19-26-16


I also get a freeze 🙄 so I'm gonna restart my computer and try again, just wanted to document :D

katelovescode:
Sometimes it's coming up partially and sometimes it's coming up fully loading - to be honest I'd rather deploy this to staging and see how it performs than trust my local.  One note is that the partially obscured horizontal scroll for the table is right next to the load bar and they look similar so I wonder if that might be confusing, but that's a @rebeccakarasiktw call I think

cjhaddad:
@katelovescode putting it on staging sounds good to me. it also could be the case that the download happens quickly and it literally goes from 0 to 100 bc the logic for reading the download sets the status of the progress bar from 0 to 100. 

I noticed that close overlap of the horizontal scroll of the table and the loading bar. can tweak padding/ margins and see how it looks better.

katelovescode (dismissed):
Yes!  Amazing!

PR #2643, opened 2022-05-26
[WIP] Adds snaplet transformations to the repo

katelovescode:


.gitignore
```diff
@@ -56,7 +56,9 @@ init.sql
 latest.dump
 
 # snaplet config
-.snaplet
+.snaplet/config.json
```
peterp, line 59: The `.snaplet/config.json` file could stay in the repo.

peterp:
If you have an existing configuration in https://app.snaplet.dev you can pull it down via `snaplet config pull`, and that may be a more accurate configuration for you personally identifiable information.

You can make modifications to that locally and then push it up via `snaplet config push`.

.gitignore
peterp: The reason why it would stay is that this is what Snaplet CLI uses to determine where your Snapshots are stored, and other people in your team will be able to run `snaplet snapshot restore` and it'll find the correct snapshots

katelovescode: Maybe my config is the old format then, because mine has my personal localhost connection string including my password in it. 🤔 I'll look at the docs and see if that should get changed.

peterp: Ah, I see! Do you use `.env` files to specify different connection strings on different machines during development? You can also use the `SNAPLET_DATABASE_URL` envar if you don't want to store the connection string:

`SNAPLET_DATABASE_URL=pg://localhost:5432/postgres snaplet proxy`

PR #2671, opened 2022-06-06
Rebase on edit message - attempt to cleanly rebase

katelovescode:


cjhaddad (changes requested):
Hi @davidcandeau I looked over the tests here and was able to quickly fix some of them. 

The only failing tests left are a couple in DashboardTitle. I believe these are failing because most if not all of the except clauses need to be wrapped in a 'waitFor'. I was looking at fixing these but I think there is a lot of logic crammed into each of these failing tests. 

Could you pull out logically chunks of logic from these two tests to make more simpler tests? This will help wrap tested logic in more easily identifiable tests/ documentation and allow us to figure out what is making these tests fail.

PR #2683, opened 2022-06-07
Pie 2632/two absences in a row

### LINKED ISSUE #2632: Unable to edit two absences to attendances in a row
## 👉🏼 Steps to Reproduce

1. Go to Edit Attendance
2. Click on an absence
3. Change it to a check in & check out time and hit save
4. Go to another absence
5. Try to change it to check in & check out time.
6. See that you are unable to make that change.

See video of the bug [here](https://drive.google.com/file/d/1wkCcv-2WmDkZ571XtYTvvS_n_TWy2Z7D/view?usp=sharing) (0:37 - 0:48) 

## 🇺🇸 States this bug is happening in
- [X] ALL
- [ ] Nebraska
- [ ] Illinois

## 🧥Expected behavior

I am able to make multiple edits in a row to absences, without having to reload the page.

## 💻 Browser Info

Please complete the following information if accessing the app from a computer:

- Operating System: [e.g. iOS]
- Browser [e.g. Chrome, Safari]

OliverCai0:
Closes #2632, which is a bug that prevents the editing of two absences in a row.

## 🧳 Steps to test
1. Go to Edit Attendance
2. Click on an absence
3. Change it to a check in & check out time and hit save
4. Go to another absence
5. Try to change it to check in & check out time.
6. See that you are unable to make that change.

## 🕯 Caveats, concerns
AttendanceDataCell to edit an attendance does not automatically get its state reset once its parent EditAttendanceModal, is hidden, so things like checkInSelected and checkOutSelected remain true after setting the check in and check out times of another attendance record. I added an effect hook that forces these variables to be reset once the record being supplied to EditAttendanceModal is changed. I think there might be a way to make this a lot more performant by removing the states in AttendanceDataCell, but that might require an overhaul of the code in there.


katelovescode:
@OliverCai0 looks great!  Please just merge or rebase from main and then you can merge this PR!

PR #2701, opened 2022-06-10
Pie 2675/generate notifications

### LINKED ISSUE #2675: Generate notifications when a child's approval is about to expire
## 📋 Summary

When any approval is within 30 days of expiration, we should generate a notification that can be accessed via the API.
When a child has a future approval that is more than 30 days expiration

- [X] ALL
- [ ] Nebraska
- [ ] Illinois

## 💻 Implementation Details

- ~~a second~~ option that's a little more fully featured:
  - add a `notifications` table where each record `belongs_to` a `child` and an `approval`
  - run a job every day to create notifications for approvals that are expiring within 30 days, where no future approval exists
  - run a job every day (maybe the same job) to clear out notifications that have been "resolved"
  - add a `notifications` endpoint (with the same policies as other controller actions)
  - add a request to the `notifications` endpoint in the notifications component already constructed
- pros: fuller functionality, can clear out notifications in the job when a child's approval is renewed in the future, performance hit of front-end parsing is eliminated
- cons: adds an additional API hit on the Dashboard, might require a loading graphic if performance is slow, this is a section of the data model we haven't incorporated in depth and might change later

per discussion with @rebeccakarasiktw and @katelovescode we want to add performance testing with large #s of children as a part of implementing this ticket

Archived notes/discussion:

> There are a couple ways we can manage this:
> - ~~the `case_list_for_dashboard` endpoint is *already* returning the information needed to generate these messages; we return `first_name` and `last_name` in the `ChildBlueprint`, and `approval_expires_on` in the nested `NebraskaDashboardCaseBlueprint`.  The front-end *could* filter the cases on the dashboard for ones expiring in the next 30 days and generate notifications at load time.~~
>   - ~~pros: no need to change the backend~~
>   - ~~cons: adds more responsibility for parsing logic on the front-end, no clear way to "clear" the messages once the child has been updated, until they switch to the new approval, possible performance hit for front-end filtering of all the cases~~

~~cc @rebeccakarasiktw and @cjhaddad - please have a look at these implementation details and let me know which of these seems like a better option (and which of the functionality I listed above is required vs. which is optional - i.e. clearing out notifications after a new approval is found)~~

~~^^ We want to go with the second option bc it allows for all criteria to be met.~~

OliverCai0:
Closes #2675. Adds notification endpoint, daily job to create and update notifications. 

## 🛷 Deployment
* [ ] Migrations
Added a migration to create notification table 20220607193259_create_notifications.rb

## 🧵 Steps to set up locally
rake daily_notifications to queue generate_notification job.

## 🧳 Steps to test
bundle exec rspec spec/services/notification_generator_spec.rb to test notification_generator behavior
bundle exec rspec spec/requests/api/v1/notifications_spec.rb  to test notifications endpoint and scope

## 🕯 Caveats, concerns
In order to achieve the desired behavior of removing notifications once an approval is updated for a child (a new approval record is created), I delete all existing notifications before querying for approvals with the latest expires_on date for each child. This seemed like the most optimal method to me, since I couldn't find a more efficient way to link a child back its latest approval other than an extensive search of all approvals, and I needed to do the query anyways to find the latest approvals.


app/blueprints/approval_blueprint.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Serializer for approvals
```
katelovescode, line 3: Let's delete this blueprint/view since we aren't using it anymore

app/blueprints/child_blueprint.rb
```diff
@@ -55,4 +55,8 @@ class ChildBlueprint < Blueprinter::Base
       )
     end
   end
+
+  view :notification do
```
katelovescode, line 59: Ditto - let's delete this view since we aren't using it anymore

app/blueprints/notification_blueprint.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+# Serializer for notifications
+class NotificationBlueprint < Blueprinter::Base
+  # association :child do |_child|
```
katelovescode, line 5: Can we remove this commented-out code?

app/services/notification_generator.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# Multiple children for expiring notification? One for each kid or a single one?
```
katelovescode, line 3: Note: we have the answer to this now - we want one notification per child, let's remove this comment.

app/services/notification_generator.rb
```diff
@@ -0,0 +1,24 @@
...
+class NotificationGenerator
+  def call
+    Notification.destroy_all
```
katelovescode, line 6: How might we do this without needing to destroy all the notification records?  What errors could this cause?

As a thought experiment:
- what happens if the job fails in the middle and the Notifications have all been destroyed but the new ones haven't finished generating?

I might suggest a different implementation, like:

- create a variable of all approvals w/ a notification i.e. `approvals_with_notification = Approval.joins(:notification)`
- loop through each of those approvals
  - if the approval is still expiring within 30 days, `return` (do nothing)
  - if not, loop through all the notifications
    - if the notification's child has no approvals after the current approval, `return` (do nothing)
    - if they do have an approval afterwards, delete the notification

- create a variable of approvals expiring within 30 days w/ no notification i.e. `approvals_without_notification = Approval.where.missing(:notification).where('expires_on < ?', 30.days.after`
- if that variable length is greater than 0, loop through each and create a new notification.

Does this make sense?  Let me know what you think.

spec/models/notification_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Notification, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
katelovescode, line 6: Let's write some specs in here!  Even though `Notification` is a tiny model, let's ensure it's properly associated.  Take a look at some of the other `spec/models` files - there will be an example of something like "`it should belong_to X`"

spec/support/api/schemas/notification.json
```diff
@@ -0,0 +1,19 @@
+{
+	"properties" : {
```
katelovescode, line 2: The spacing on this JSON is very large, can you make sure you're only using 2 spaces instead of tabs/4+ spaces?

spec/support/api/schemas/notifications.json
```diff
@@ -0,0 +1,6 @@
+{
```
katelovescode, line 1: Ditto spacing comment here

app/models/notification.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class Notification < ApplicationRecord
+  belongs_to :child
+  belongs_to :approval
+  validates :approval_id, uniqueness: { scope: :child_id }
```
katelovescode, line 6: Great uniqueness validation here, I hadn't thought about this

config/routes.rb
```diff
@@ -31,6 +31,7 @@
       resources :attendances, only: %i[index update destroy]
       resources :service_days, only: %i[index create]
       resources :attendance_batches, only: :create
+      resources :notifications, only: :index
```
katelovescode, line 34: Glad this is scoped only to exactly what we need!

db/seeds.rb
```diff
@@ -178,6 +179,16 @@ def create_case(first_name:,
     )
   end
 
+  if add_expiring_approval
```
katelovescode, line 182: Love this addition to the seeds, thank you!

katelovescode (changes requested):
Really great work, this is a *huge* feature to work on right out of the gate!  I added a couple of suggestions on implementation, and a few requests for some tweaks.  Happy to talk about this on Monday!

app/services/notification_generator.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+# Creates notifications for each approval that is expiring within 30 days. Clears
+# expiring notifications and any notifications that have received updated approvals
+class NotificationGenerator
+  def call
+    approvals_without_notification = Approval.where.missing(:notifications).where('expires_on between ? and ?',
```
katelovescode, line 7: Handy syntax here:
```
Approval.where.missing(:notifications).where(expires_on: 0.days.after..30.days.after)
```
the range notation `..` works for dates

app/models/child.rb
```diff
@@ -16,6 +16,7 @@ class Child < UuidApplicationRecord
   has_many :service_days, dependent: :destroy
   has_many :attendances, through: :service_days, dependent: :destroy
   has_many :nebraska_dashboard_cases, dependent: :destroy
+  has_one :notification, dependent: :destroy
```
katelovescode, line 19: 😍

app/models/child_approval.rb
```diff
@@ -9,6 +9,7 @@ class ChildApproval < UuidApplicationRecord
   has_many :nebraska_approval_amounts, dependent: :destroy
   has_many :attendances, dependent: :destroy
   has_many :service_days, -> { distinct }, through: :attendances
+  has_one :notification, dependent: :destroy
```
katelovescode, line 12: Out of curiosity, why are you also associating to a child_approval?  Was this a mistake?  I ask because I don't see you using this association anywhere.

app/models/notification.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class Notification < ApplicationRecord
+  belongs_to :child
+  belongs_to :approval
+  # validates :approval_id, uniqueness: { scope: :child_id }
```
katelovescode, line 6: How come you decided against this scoped uniqueness?

app/services/notification_generator.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+    end
+
+    delete_notifications
+  end
+
+  private
+
+  def delete_notifications
+    Approval.all.each do |record|
```
katelovescode, line 22: Rather than looping through all Approvals, does it make sense to reduce the number of Approvals you're checking with a join?  `Approval.joins(:notifications).each`?

katelovescode, line 22: Rather than `record` can we name the param `approval` so it's clear what we're operating on?

app/services/notification_generator.rb
```diff
@@ -0,0 +1,36 @@
...
+
+# Creates notifications for each approval that is expiring within 30 days. Clears
+# expiring notifications and any notifications that have received updated approvals
+class NotificationGenerator
+  def call
+    approvals_without_notification = Approval.where.missing(:notifications).where('expires_on between ? and ?',
+                                                                                  0.days.after,
+                                                                                  30.days.after)
+    if approvals_without_notification.length.positive?
+      approvals_without_notification.each do |record|
```
katelovescode, line 11: Can we name this param `approval` rather than `record` for clarity?

app/models/notification.rb
OliverCai0: I'm not sure if it is still necessary given that there should be only one notification per child since there is only one approval per child.

app/services/notification_generator.rb
OliverCai0: I wasn't sure if the updated approvals shared the same ID. The approvals that I utilized in the test didn't share id and this join was problematic in the fact that the preexisting notification always linked back to the non updated approval, and I needed to check all the most recent approvals.

app/models/child_approval.rb
OliverCai0: Yes this was a mistake.

PR #2748, opened 2022-06-24
Pie 2693/new attendance type

### LINKED ISSUE #2692: Remove validation that requires ServiceDay w/ an absence_type to be associated to a Schedule
## 📋 Summary

We are going to allow absences to be created on days without a Schedule, so everywhere we've restricted ServiceDay/absence creation based on if the child is scheduled or not will need to be removed.

BLOCKED BY #2693

### LINKED ISSUE #2693: Add new absence/attendance types and update logic
## 📋 Summary

- Add `absence_on_scheduled_day` to allowed absence types
- Add `absence_on_unscheduled_day` to allowed absence types
- Update all logic (calculations for the dashboard) to include both of these new types as if they were regular `absence`s
- Change all existing absences to `absence_on_scheduled_day` given prior restrictions to be removed in #2692 
- Change incoming ServiceDay#creates and #updates from AttendanceDay to take `absence` from the `absence_type` value and determine if it is scheduled or unscheduled, and assign the appropriate type (this will eventually be moved to a command in #2570 and #2571 )
- Document the API: incoming values can be `absence` or `covid_absence` or `nil` for `ServiceDay.absence_type` - the backend will decide if it's scheduled or not

Doc for context: https://github.com/pieforproviders/pieforproviders/blob/main/docs/attendances_and_absences.md

OliverCai0:
Closes #2693 
Closes #2692

### DIFF
--- a/spec/models/service_day_spec.rb
+++ b/spec/models/service_day_spec.rb
@@ -122,12 +122,12 @@ RSpec.describe ServiceDay, type: :model do
         :service_day,
         child: child,
         absence_type: 'covid_absence',
-        date: Helpers.prior_weekday(Time.current, 1).in_time_zone(child.timezone).at_beginning_of_day
+        date: Helpers.prior_weekday(Time.current, 2).in_time_zone(child.timezone).at_beginning_of_day
       )
     end
 
     let!(:service_day_with_attendance) { create(:service_day) }
-    let(:attendance) { create(:attendance, service_day: service_day) }
+    let(:attendance) { create(:attendance, service_day: service_day_with_attendance) }
 
     it 'returns absences only' do
       expect(described_class.absences).to include(absence)


katelovescode (changes requested):
We should do a deeper dive on this, I'm going to add it to the team sync discussion today because there's undefined behavior.

katelovescode (changes requested):
Couple more requests for change, we can talk about this a little later today.  Great job!

spec/models/service_day_spec.rb
```diff
@@ -87,6 +138,9 @@
     end
 
     it 'returns standard absences only' do
+      absence.reload
+      child.reload
+      # binding.pry()
```
katelovescode, line 143: Let's remove this!

spec/requests/api/v1/service_days_spec.rb
```diff
@@ -717,7 +717,7 @@
         post '/api/v1/service_days', params: params_with_absence, headers: headers
         parsed_response = JSON.parse(response.body)
         expect(parsed_response['child_id']).to eq(child.id)
-        expect(parsed_response['absence_type']).to eq(params_with_absence[:service_day][:absence_type])
+        expect(parsed_response['absence_type']).to be_in(%w[absence_on_scheduled_day absence_on_unscheduled_day])
```
katelovescode, line 720: Can we fix this one to be deterministic as well?

katelovescode (dismissed):

PR #2763, opened 2022-06-28
Pie 2694/user authorization and policy scoping

### LINKED ISSUE #2694: Ensure consistent user authorization and policy scoping for Attendances and ServiceDays
## 📋 Summary

This could be a PR for all controller actions; users should be authorized for all internal API endpoints (not the ones outside the API like registration and password).

All API endpoints should have:
- before_action to `set_[model]` for the `update` and `destroy` actions scoped using Pundit
- before_action to `set_[models]` for the `index` action scoped using Pundit
- before_action to authorize the user

We also want to write tests for checking authorizations

OliverCai0:
Closes #2694


app/models/service_day.rb
```diff
@@ -23,6 +23,7 @@ class ServiceDay < UuidApplicationRecord
   validates :absence_type, inclusion: { in: ABSENCE_TYPES }, allow_nil: true
   validates :date, date_time_param: true, presence: true
   validate :prevent_creation_of_absence_without_schedule
+  validates :child, uniqueness: { scope: :date }
```
katelovescode, line 26: Is this pulled in from the other PR for uniqueness indexing?  If so, we should remove it from this PR

app/controllers/api/v1/service_days_controller.rb
```diff
@@ -5,12 +5,32 @@ module V1
...
+        else
+          render json: @service_day.errors, status: :unprocessable_entity
+        end
+      end
+
+      # DELETE /service_days
+      def destroy
+        result = @service_day.destroy
+        if result
+          render json: result, status: :destroyed
```
katelovescode, line 28: If I remember correctly, the better response for destruction is `no_content` - check out how we're doing `destroy` in the AttendancesController for guidance here (it's using a command, so you could do:

```
      def destroy
        if @service_day
          @service_day.destroy
          render status: :no_content
        else
          render status: :not_found
        end
      end
```

db/migrate/20220621165259_add_unique_key_to_service_day.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueKeyToServiceDay < ActiveRecord::Migration[6.1]
+  def change
+    add_index :service_days, %i[child_id date], unique: true
```
katelovescode, line 3: Ditto my above comment, this is from the other PR, right?

db/schema.rb
```diff
@@ -10,10 +10,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2022_06_08_035549) do
+ActiveRecord::Schema.define(version: 2022_06_21_165259) do
```
katelovescode, line 13: Won't need these schema changes, as well, given that there's no DB change in this PR

spec/factories/service_days.rb
```diff
@@ -40,9 +40,10 @@
 #
 # Indexes
 #
-#  index_service_days_on_child_id     (child_id)
```
katelovescode, line 43: Also remove this since it's from the other PR.

spec/models/child_spec.rb
```diff
@@ -173,12 +173,16 @@
       expired_child_approval = expired_approval.child_approvals.where(child: child).first
       current_attendances = []
       expired_attendances = []
-      3.times do
-        service_day = create(:service_day, child: child)
+      3.times do |idx|
```
katelovescode, line 176: Also remove this since it's from the other PR.

spec/models/service_day_spec.rb
```diff
@@ -13,6 +13,10 @@
   it { is_expected.to belong_to(:child) }
   it { is_expected.to validate_presence_of(:date) }
 
+  it 'validates the uniqueness of child_id scoped to date' do
```
katelovescode, line 16: Also remove this since it's from the other PR.

katelovescode (changes requested):
This sets up the before actions to do policy scope but we're missing authorization - check out `businesses_controller.rb`:

```
before_action :authorize_user, only: %i[show update destroy]
```

We'll want to add this auth action to all API endpoints (attendance batches, attendances, notifications, and service days are missing them)

==================

Can you also:
- update the businesses_controller index method to use the `before_action :set_businesses` like you did in Attendances
- add missing endpoints to the other controllers (update, destroy, etc.) - check all the controllers and see which CRUD actions are missing and fill them in, please!

app/controllers/api/v1/notifications_controller.rb
```diff
@@ -4,19 +4,46 @@ module Api
...
+      before_action :set_notification,  only: %i[update destroy]
       before_action :set_notifications, only: %i[index]
 
       def index
         render json: NotificationBlueprint.render(@notifications)
       end
 
+      def update
+        if @notification
+          @notification.update(notification_params)
```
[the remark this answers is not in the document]
OliverCai0: The extra authorize in the index action of Users seems to be necessary after testing.

app/controllers/api/v1/users_controller.rb
```diff
@@ -5,11 +5,10 @@ module V1
     # API for application users
     class UsersController < Api::V1::ApiController
       before_action :set_user, only: %i[show]
-      before_action :authorize_user, only: %i[show]
+      before_action :set_users, only: %i[index]
 
       # GET /users
       def index
-        @users = policy_scope(User)
         authorize User
```
OliverCai0, line 12: Several tests fail after removing this authorize method, so it might be better to keep it rather than relying on just the set_users

app/services/commands/notification/create.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Commands
```
katelovescode, line 3: I've been writing Commands for ServiceDays and attendances because they're complex and have lots of callback behavior I'd like to call explicitly.  I don't think it's necessary to have a create command for notifications.

app/services/notification_generator.rb
```diff
@@ -7,6 +7,10 @@ def call
     sync_notifications
   end
 
+  def singular_call(child, approval)
```
katelovescode, line 10: Can you tell me more about this method and what you're using it for?

spec/support/api/schemas/service_day.json
```diff
@@ -30,6 +30,7 @@
     },
     "total_time_in_care": {
       "type": [
+        "integer",
```
katelovescode, line 33: I don't think this should be necessary; can we walk through the spec that's throwing this error together?  I've seen it before and it's usually a symptom of another problem.

katelovescode (commented):
Notes:
- user is still missing create, update and destroy
- want to dig in a little more on the policy_scope - if you take a look at ApplicationPolicy we have some methods defined like `create?` and `index?` and then in the UserPolicy we're overwriting `index?` and `create?` but maybe that's not working?  As you can see this is *old* code we haven't touched in a while so it could use some research/digging.  Maybe there's a better way to handle it.

app/controllers/api/v1/users_controller.rb
katelovescode: See my comment below - I think we have some conflicting behavior in the `ApplicationPolicy` that every policy inherits from and the `UserPolicy` - we might have to explicitly call authorize in `index` in order to get the `UserPolicy.index?` method to be triggered.  Would love to think about that and maybe specify this behavior in all other models as well?  We should chat about this, maybe later today or next week.

app/services/notification_generator.rb
OliverCai0: This is something I accidentally left over. I've removed it.

.rubocop.yml
```diff
@@ -26,6 +26,10 @@ Metrics/ClassLength:
 
 Metrics/MethodLength:
   CountAsOne: ['array', 'heredoc', 'hash']
+  Exclude:
+    - app/controllers/api/v1/users_controller.rb
```
OliverCai0, line 30: I wasn't able to find an easy way to minimize the number of lines in the user_params of the User Controller, so I added it to be excluded here. Upon adding it, the methodlength warning went off for several additional controllers, so I added them as well.

spec/support/api/schemas/service_day.json
```diff
@@ -30,6 +30,7 @@
     },
     "total_time_in_care": {
       "type": [
+        // "integer",
```
katelovescode, line 33: @OliverCai0 you'll need to remove this - you can't comment in JSON :)

katelovescode (approved):

Issue #2577, opened 2022-05-06
Refactor the Wonderschool CSV import


katelovescode:
## 📋 Summary

Refactor the Wonderschool CSV: instead of processing every row as-is, group by day, do some validation on the data, throw out any attendances that don't need to be touched, make new ones, make edits, etc.

Rows that should be thrown out before attempting to process:
- kids that are not in our system using the Child -> Wonderschool ID
- attendances whose unique ID is already in our system using the Attendance -> Wonderschool ID, where the rest of the values are the same (check_in, check_out, absence) **NOTE** the absence flag now lives on ServiceDay so make sure you're checking the right model
- (TBD) Rows outside the date range?  See implementation details below.

Rows that should create a new record:
- a unique Wonderschool ID we've never seen

Rows that should be used to update our records:
- existing Wonderschool ID with new data in check_in, check_out, and/or absence **NOTE** the absence flag now lives on ServiceDay so make sure you're checking the right model

Rows that should be used to delete our records (MAYBE):
- Wonderschool IDs that are *missing* from the CSV result but we have in our system.  @rebeccakarasiktw I'd love to talk more about this at another backlog refinement (this is not work that has to be done before 5/10, so it can wait)

## 🇺🇸 States this change applies to
- [X] ALL
- [ ] Nebraska
- [ ] Illinois

## 💻 Implementation Details

Possibly group by date and assume we're replacing the whole data set - from my convo w/ Andy Nu:
> Trust the data from the 3rd party. If they are giving you data for a date range you are replacing the entries in your system entirely for that range for the Attendance records, and then your cron will come along and deal with flipping the absence on the ServiceDay if it needs it after the fact. Just make sure that your cron is paying attention to the same time period that your 3rd party data pulls are covering.

PR #2802, opened 2022-07-14
PIE-2577/wonderschool date range

OliverCai0:
Addresses the date range feature specified in #2577 

## 🕯 Caveats, concerns
Date range created by environment variables set in the rake task command. No variable input ensures original behavior.


app/services/attendance_csv_importer.rb
```diff
@@ -40,6 +42,8 @@ def process_row(row)
     @row = row
     @child = child
 
+    return unless (@start_date..@end_date).cover?(@row['check_in'].in_time_zone(@child.timezone).at_beginning_of_day)
```
OliverCai0, line 45: Date range condition applied here in order to handle the timezone. Can be applied earlier in the process row block, but the child wouldn't necessarily exist and then we wouldn't be able to find the appropriate timezone which could be relevant.

lib/tasks/attendance_csv_importer.rake
```diff
@@ -4,7 +4,9 @@ task attendance_csv_importer: :environment do
   # Pull single file from Wonderschool URL
   # Process single file through Attendance Processor
   # Archive file to S3
-  AttendanceCsvImporter.new.call
+  start_date = ENV['start_date']&.to_date || 1.year.before
```
katelovescode, line 7: Can you talk to me about the choice to pass a static date or default to a dynamic time range here?

app/services/attendance_csv_importer.rb
katelovescode: This is definitely the right place to do this!

app/services/attendance_csv_importer.rb
```diff
@@ -17,7 +17,9 @@ def initialize
     @archive_bucket = Rails.application.config.aws_necc_attendance_archive_bucket
   end
 
-  def call
+  def call(start_date = nil, end_date = 0.days.after)
```
katelovescode, line 20: rather than set these in the call block, I'd rather we do it in the initialize and leave the call block pristine, so it'd be like AttendanceCsvImporter.new(start_date, end_date).call

Also - can you use named params instead of positional ones?  (i.e. `def initialize(start_date: nil, end_date: 0.days.after)`

katelovescode (commented):
A couple of notes on tests and using the initialize block, and then an open question about the way you implemented.  Let me know what you think!

lib/tasks/attendance_csv_importer.rake
OliverCai0: I'm still not exactly sure about what the optimal date range is, so I wanted to provide the option to configure the range according to the center. I'm assuming that without params, setting the earliest date to a year ago is enough to catch all the changes. This is all assuming that this task is used on a regular basis with most recent dates in mind, if this task is more often used to onboard the entire attendance records of a single new center, I believe it might be better to just default to nil, so that the default behavior is to look at all records.

katelovescode: Ahhhh gotcha!  So, for more clarity, this task is run every single day to pull data from Wonderschool.  Wonderschool provides us with a CSV link to *every attendance that has ever been entered* for all of the users in our system, and we pull that every day to get new attendance records.  What changes would you make based on that information?

OliverCai0: In that case I can default the start date value to the day before?

rebeccakarasiktw: If a year ago is doable without impacting performance, a year would be great. If not, one month back is good.

katelovescode: @OliverCai0 I think a year is reasonable, shouldn't be too bad on performance and we can always change it later.  Can you set it to 1.year.ago from the date the task is run?

lib/tasks/attendance_csv_importer.rake
```diff
@@ -4,7 +4,9 @@ task attendance_csv_importer: :environment do
...
+  end_date = ENV['end_date']&.to_date || 0.days.after
+  AttendanceCsvImporter.new.call(start_date, end_date)
```
katelovescode, line 9: Don't forget to update this implementation to initialize with the named variables!

app/services/attendance_csv_importer.rb
```diff
@@ -11,13 +11,15 @@ class NoSuchBusiness < StandardError; end
...
+  def initialize(start_date: nil, end_date: 0.days.after)
     @client = AwsClient.new
     @source_bucket = Rails.application.config.aws_necc_attendance_bucket
     @archive_bucket = Rails.application.config.aws_necc_attendance_archive_bucket
+    @start_date = start_date&.at_beginning_of_day
+    @end_date = end_date&.at_end_of_day
   end
 
-  def call
+  def call(_start_date = nil, _end_date = 0.days.after)
```
katelovescode, line 22: Don't forget to remove these unnecessary params now :)

lib/tasks/attendance_csv_importer.rake
katelovescode, line 7: Can we rename these env vars to something like `wonderschool_attendance_import_start_date` so that it's clear when configuring what the start date/end date are for?

katelovescode (changes requested):
Small change request, looks great though!

lib/tasks/attendance_csv_importer.rake
```diff
@@ -4,7 +4,9 @@ task attendance_csv_importer: :environment do
   # Pull single file from Wonderschool URL
   # Process single file through Attendance Processor
   # Archive file to S3
-  AttendanceCsvImporter.new.call
+  start_date = ENV['wonderschool_attendance_import_start_date']&.to_date || 1.year.before
```
katelovescode, line 7: oh shoot!  I forgot here:
Can you document this env var in the `.env.sample` file? And also our standard for env vars is to do them in all caps, can you change these two to all caps here as well?

katelovescode (changes requested):

katelovescode (approved):

PR #2824, opened 2022-07-22
PIE-2819/delete_by_user_email

OliverCai0:
Provides a reproducible method to delete users and their associated records by email, such as in #2819.


db/migrate/20220722155443_add_cascade_to_user_associated_records.rb
```diff
@@ -0,0 +1,8 @@
+class AddCascadeToUserAssociatedRecords < ActiveRecord::Migration[6.1]
+  def change
+    remove_foreign_key :businesses, :users
+    add_foreign_key :businesses, :users, on_delete: :cascade
```
katelovescode, line 4: We might not need this either if you get rid of the restrict_with_error on all models.

spec/models/user_spec.rb
```diff
@@ -51,6 +51,47 @@
     expect(new_user.errors.messages[:email]).to include('has already been taken')
   end
 
+  describe '#destroy' do
```
katelovescode, line 54: These tests are probably not necessary because they're testing the default behavior of Rails, including the "delete cascading" - if it was me, I would remove these tests (and the restrict_with_errors as I mentioned above) and assume Rails will handle this correctly.

katelovescode (changes requested):
Few changes, but let me know what you think!

katelovescode (approved):

PR #3207, opened 2023-03-30
Pie 3130 upgrade Ant Design version

MartinM1312:
Ant Design version upgraded

## 🍂 Before You Submit

## 🛷 Deployment

* [ ] Migrations
* [ ] Dependencies


client/src/Dashboard/DatePicker.js
```diff
@@ -1,6 +1,7 @@
 import dayjsGenerateConfig from 'rc-picker/lib/generate/dayjs'
 import generatePicker from 'antd/lib/date-picker/generatePicker'
-import 'antd/lib/date-picker/style/index'
+
+// import 'antd/lib/date-picker/style/index'
```
fabianraf, line 4: If not needed, remove it.

client/src/_shared/Header.js
```diff
@@ -90,6 +91,7 @@ export function Header() {
                 className={
                   location.pathname.includes('/attendance') ? 'underline' : ''
                 }
+                id={'dashboard-tab'}
```
fabianraf, line 94: You cannot have the same id twice in a form

fabianraf (changes requested):

client/src/Dashboard/DatePicker.js
MartinM1312: done

client/src/_shared/Header.js
MartinM1312: fixed

